PR #1, opened 2019-05-19
support custom/non-global fetch implementation

Xiphe:
in order not not rely on a globally available fetch, support easier mocking, etc.


fgnass:
+1 for the feature but the compilation now fails with a `ReferenceError`.

Xiphe:
fixed the reference error. Now the `should place orders` test fails. But also on master. Could you check that @fgnass?

fgnass:
http://petstore.swagger.io/#/store/placeOrder always returns 500 at the moment. I guess we should skip our test for now or invest some time to mock the petstore api.

Xiphe:
@fgnass PTAL 🚀

fgnass:
oazapfts@1.1.0

PR #3, opened 2019-05-27
feat(generate): create Server classes from API servers

Xiphe:
in order to expose all potential URLs and variables of the apis servers.

Given swagger spec:

```json
{
  "servers": [
    {
      "description": "MyApi",
      "url": "https://petstore.swagger.io/{version}",
      "variables": {
        "version": {
          "enum": ["v1", "v2"],
          "default": "v2"
        }
      }
    }
  ]
}
```

Usage:

```ts
import { Api, MyApiServer } from './Api.ts';

const api = new Api({ server: new MyApiServer({ version: "v1" }) });
```

I deprecated `ApiOptions["baseUrl"]` in favour of `ApiOptions["server"]` which can either be a string or an object with url prop.

PTAL: @fgnass

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,1 @@
+{ "editor.formatOnSave": false }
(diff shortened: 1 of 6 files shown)


dlehmhus:
LGTM

Issue #14, opened 2020-04-15
Contributing


rtfpessoa:
I stumbled on the oazapfts (https://github.com/cellular/oazapfts) library recently, while I was in the beginning of my own generator.
I found that the ideas are very aligned with mine and the code is very approachable.

There are some features I would like to have, like multiples files output, separate model generation from the actual endpoint call, wrappers for endpoint responses, payload validation, ...

I would like to know if you have any interest in receiving contributions? and If you have any other kind of plans and missing features list for the library?

fgnass:
We (at least for now) consider the library to be feature complete but are open for ideas and PRs. I would like to learn about the rationale behind the features you mentioned:

* _multiples files output_: What would go into the different files? How would you like to see things being split (and why)?
* _separate model generation_: You mean actual classes? Would they contain any logic or would they be plain data transfer objects? 
* _wrappers for endpoint responses_: What would these wrappers do? 
* _payload validation_: Request or response validation (or both)?

rtfpessoa:
Thanks for the quick answer.
We (at least for now) consider the library to be feature complete but are open for ideas and PRs. I would like to learn about the rationale behind the features you mentioned:

> multiples files output: What would go into the different files? How would you like to see things being split (and why)?
The idea would be to have the models separate from the actual endpoints implementation and then, since some apis can grow a lot, we could have one file per tag in the swagger.

> separate model generation: You mean actual classes? Would they contain any logic or would they be plain data transfer objects?
Just play interfaces/types but separate from the implementation. You could even just generate the models and not the implementation in some cases.

> Wrappers for endpoint responses: What would these wrappers do?
In some cases where we do not have a schema defined inside components for a response it ends up being an inline struct. If you created something like a named type for each response it could even be referenced in the rest of the code.
e.g.:
```typescript
type myOperationIdResponse = myOperationIdResponse200 | myOperationIdResponse404
type myOperationIdResponse200 = {
  foo: string
}
type myOperationIdResponse404 = Error
type Error = {
  message: string
}
```

> payload validation: Request or response validation (or both)?
I was just considering response. Specifically two parts. We could always validate if the status codes returned were in the spec (or if a default is defined) and if it is a json media we could validate the schema and reject the promise if it does not match.
Ideally I would always resolve promises that are specified in the spec and the response would be a union type.

Let me know what you think of each of them.

fgnass:
> > multiples files output: What would go into the different files? How would you like to see things being split (and why)?

[...]

PR #15, opened 2020-04-17
chore: Add prettier configuration

rtfpessoa:
While I was looking into the details discussed in #14 I had some issues with formatting.

Since I use VSCode the easiest way to format code is through prettier.

I tried to replicate the style in the project in one configuration and ended up with this.

What do you think? Are you interested in having it setup in the CI and/or having a [husky](https://www.npmjs.com/package/husky) hook?

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ The generated file exports a `defaults` constant that can be used to override th
 import * as api from "./api.ts";
 
 api.defaults.headers = {
-  access_token: "secret"
+  access_token: "secret",
 };
 ```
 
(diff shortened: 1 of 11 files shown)


fgnass:
We already have a prettier config in [package.json](https://github.com/cellular/oazapfts/blob/master/package.json#L57). I had to add it recently when prettier changed their default settings and I didn't want to reformat the whole project while working on a feature, so this was a quick fix.

I would prefer to add prettier as devDependency and remove the config altogether so that we can stick to the prettier defaults. I think running prettier as eslint-plugin and setting up husky to run `eslint --fix` would be the best approach.

rtfpessoa:
Followed the guide here https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged

I guess eslint could be a next step but would probably involve a bit more work

PR #16, opened 2020-04-17
feat: Resolve recursive references

rtfpessoa:
One feature I forgot to mention before, that was fairly simple to implement was recursive references.

### DIFF
--- a/src/generate.ts
+++ b/src/generate.ts
@@ -180,14 +180,18 @@ function supportDeepObjects(params: OpenAPIV3.ParameterObject[]) {
 export default function generateApi(spec: OpenAPIV3.Document) {
   const aliases: ts.TypeAliasDeclaration[] = [];
 
-  function resolve<T>(obj: T | OpenAPIV3.ReferenceObject) {
-    if (!isReference(obj)) return obj;
-    const ref = obj.$ref;
-    if (!ref.startsWith("#/")) {
-      throw new Error(`External refs are not supported: ${ref}`);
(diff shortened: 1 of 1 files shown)


src/generate.ts
```diff
@@ -180,14 +180,18 @@ function supportDeepObjects(params: OpenAPIV3.ParameterObject[]) {
 export default function generateApi(spec: OpenAPIV3.Document) {
   const aliases: ts.TypeAliasDeclaration[] = [];
 
-  function resolve<T>(obj: T | OpenAPIV3.ReferenceObject) {
-    if (!isReference(obj)) return obj;
-    const ref = obj.$ref;
-    if (!ref.startsWith("#/")) {
-      throw new Error(`External refs are not supported: ${ref}`);
+  function resolve<T>(obj: T | OpenAPIV3.ReferenceObject): T {
+    function resolveAux<T>(obj: T | OpenAPIV3.ReferenceObject): T {
```
fgnass, line 184: Why do we need this inner function? Couldn't `resolve` call itself in line 191?

fgnass:
Thank you! Can you add a test case for recursive references?

src/generate.ts
rtfpessoa: You are completely right.

rtfpessoa:
Forget about this, it is already supported.
I mistakenly tough this would be needed but it is not, since I need something similar for the validation, but while doing the test I might found a better way of doing that.
Sorry for the confusion.

PR #17, opened 2020-04-17
fix: Use process exitCode instead of exit

rtfpessoa:
One small detail I noticed and tough could be interesting to tweak was the exit usage. 

Check https://eslint.org/docs/rules/no-process-exit for more details.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -13,7 +13,7 @@ const [, , spec, dest] = process.argv;
 
 if (!spec) {
   console.error("Usage: oazapfts <spec> [filename]");
-  process.exit(1);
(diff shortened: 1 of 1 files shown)


fgnass:
Hm, I disagree with eslint's reasoning here. I want the process to immediately exit in that line so that future code we might add below doesn't get executed by mistake. This isn't some code deeply nested inside an application but the main entry file that is responsible for managing the process lifecycle.

rtfpessoa:
Fair enough.
I guess my view here is mostly from previous experience where I did not knew about this, and eventually killed the process in a bad place.

PR #19, opened 2020-04-24
Validate responses

rtfpessoa:
- Validate response body schemas if in json
- Validate status codes for other encodings

@fgnass this is some initial work for validating the schemas.
Can you take a quick look to see if it would be in line with our discussion?

It does not contain the optional/configurable part, but I would like to have a nice working version before going for that.

### DIFF
--- a/demo/api.test.ts
+++ b/demo/api.test.ts
@@ -22,10 +22,15 @@ describe("petstore.swagger.io", () => {
     expect(pet2).toMatchObject(pet);
   });
 
-  it("should reject invalid ids", () => {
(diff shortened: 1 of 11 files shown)


fgnass:
Before we add validation lets discuss how we want to handle responses in future. Please chime in at #20

fgnass:
FIY: I created a `next` branch that now resolves all responses (not only the ones with a status < 400) like you suggested and also moved the runtime to a separate file. See https://github.com/cellular/oazapfts/issues/20#issuecomment-621488128 for details.

Issue #21, opened 2020-05-05
Support for external refs?


dradetsky:
Do you have any plans to add support for external refs soon? That is, a definition of the form

```
MyObjectName:
  $ref: 'path/to/file.yaml#path/to/MyObjectName'
```

Background:

My company is making a lot of microservices. I'm trying to introduce schema-first design for these, and ultimately hoping to add utilities for simplifying this to our company typescript library. Naturally, I want to get all of the value I can out of the schema, like generating typescript types. However, I strongly suspect I'll ultimately need to pull common bits of the schema (like the types of error responses) into a library so they can be standard across many services.

I considered trying to add external ref support myself. I know that [openapi-backend](https://github.com/anttiviljami/openapi-backend) supports external refs, and they use the same parser as oazapfts, so I figure I could copy what they did. However:

0. I'm fairly busy, so I probably won't get to it very quickly.

1. I don't want to do it if you're already halfway done with it, because you'll probably do a better job.

fgnass:
I think the basic idea would be to turn external refs into `import MyObjectName from "path/to/file.ts"`. Is it safe to assume that the relative paths of the API specs are the same as the ones of the generated TypeScript files? Otherwise we would need some form of mapping config.

dradetsky:
@fgnass It sounds like instead of actually resolving external refs, you want to just recursively invoke the oazapfts cli on all the files named in the external refs, is that right?

If so, one issue with this is that you can't actually just invoke the cli on a file which is a valid external ref, but not an actual complete openapi definition. For example I can do

```
# file0.yaml
components:
  schemas:
    MyObj:
      $ref: 'file1.yaml#/MyObj'
---
# file1.yaml
MyObj:
  type: object
  ...
```

And this is valid openapi, but the oazapfts cli can't be invoked on file1 because it's not a valid openapi spec, just a valid fragment of one. Maybe this is possible to do by using just the various internal functions of oazapfts instead of the cli or main entry point, but I don't actually know the code yet.

The other point is that if you look at [openapi-backend](https://github.com/anttiviljami/openapi-backend/blob/master/src/backend.ts#L180) you'll see that it looks like they solve this with just one more call (`SwaggerParser.dereference`) to the parser library you both share. Maybe there's more to it than that, but I figure that at least you might be able to tell immediately that it wouldn't just work.

dradetsky:
@fgnass i think i figured it out, pr incoming

fgnass:
[...]

PR #23, opened 2020-05-06
Allow external refs

dradetsky:
PR for issue #21 

In theory this is just the tiny change from `parse` to `bundle`. It appears that this is what we want to do from [their documentation](https://apitools.dev/swagger-parser/docs/swagger-parser.html#bundleapi-options-callback):

> Bundles all referenced files/URLs into a single api that only has internal $ref pointers. This lets you split-up your API however you want while you’re building it, but easily combine all those files together when it’s time to package or distribute the API to other people. The resulting API size will be small, since it will still contain internal JSON references rather than being fully-dereferenced.

I've also included the test cases I used to implement them. As you see, there are 2 yaml files, one of which has 1 fragment moved to a third yaml file. The resulting output was identical.

I would have liked to add a test to verify this, although I had a small problem. When I ran tests, it modified `demo/api.ts` such that the indentation was increase from 2 spaces to 4. I don't know why this happened. It wasn't related to this change; running the tests on a clean repo (deps installed with `npm ci`) made the same change on my machine.

### DIFF
--- a/demo/extern.yaml
+++ b/demo/extern.yaml
@@ -1,0 +1,14 @@
+PartnerInfo:
+  type: object
+  required:
+    - name
+  properties:
+    name:
+      type: string
+    flavor:
+      type: string
+    bestBeforeDate:
+      type: string
+      format: date
+    numInPack:
+      type: integer
--- a/demo/full.yaml
+++ b/demo/full.yaml
@@ -1,0 +1,55 @@
+openapi: 3.0.1
+
+info:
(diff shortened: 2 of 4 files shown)


dradetsky:
By the way, feel free to cherry-pick just the code change. We don't (currently) need those test files. I just stuck them in because they were there & might help.

fgnass:
Fix landed in `oazapfts@3.0.0-rc4`

PR #36, opened 2020-11-04
docs: omit reference to official tooling from README.md

MikeRalphson:
There is no official OpenAPI tooling (though there are two popular open-source Java-based tools).


MikeRalphson:
Thanks! And I see you're using swagger2openapi, cool. Will check out this project further.

MikeRalphson:
If you add the topic `openapi3` to the repo, it will automatically be included in https://apis.guru/awesome-openapi3/

PR #43, opened 2020-12-12
Convert generateApi to class, upgrade to TS 4.1.2, add exports

phryneas:
So, as already announced in https://github.com/cellular/oazapfts/issues/41#issuecomment-739527701, this PR would 
* convert the `generateApi` function to a class & expose it's inner functions, 
* add exports for a lot of functions 
* upgrade TS to version 4.1.2, using the new `factory` (`ts.create...` is deprecated now)

AST element properties are now marked as `readonly` and in some places they were deliberately modified. Since it didn't seem to cause any problems, I now just re-assign them with `Object.assign`, but that might break in a future TS version.

### DIFF
--- a/demo/api.test.ts
+++ b/demo/api.test.ts
@@ -16,9 +16,9 @@ describe("ok", () => {
 
   it("should throw if status != 200", async () => {
     const promise = ok(
-      api.getPetById(1, { headers: { Prefer: "statusCode=404" } })
+      api.getPetById(4, { headers: { Prefer: "statusCode=404" } })
     );
-    expect(promise).rejects.toHaveProperty("status", 404);
+    await expect(promise).rejects.toHaveProperty("status", 404);
   });
 
   it("should post json", async () => {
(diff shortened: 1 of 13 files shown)


kahirokunn:
@benjidea How about this one?

fgnass:
@phryneas I'm so sorry, I was busy with other stuff and totally forgot about this PR! That being said, would you be interested in becoming an oazapfts maintainer?

phryneas:
It is a tempting offer and thank you for the trust - but I already have too much on my plate with the whole Redux Toolkit + surrounding ecosystem, so unfortunately I won't have the time for that :slightly_smiling_face: 

Need me to fi up the conflicts on this one?

fgnass:
That would be awesome if you could take a look at the conflicts, I started merging them but gave up on the halfway through.

phryneas:
Okay, I've re-applied all changes to `generate.ts` manually and then merged the rest, I think this should be fine. Tests still work.

Issue #40, opened 2020-11-17
Types are generated with spaces in the names when spec has titles with spaces


rtfpessoa:
This can be replicated by running the CLI over 
https://raw.githubusercontent.com/OpenAPITools/openapi-generator/master/modules/openapi-generator/src/test/resources/3_0/petstore.yaml

which will generate types like:

```ts
export type A Pet = {
    id?: number;
    category?: Pet category;
    name: string;
    photoUrls: string[];
    tags?: Pet Tag[];
    status?: "available" | "pending" | "sold";
};
export type An uploaded response = {
    code?: number;
    "type"?: string;
    message?: string;
};
```

gjsduarte:
The issue seems to be coming from https://github.com/cellular/oazapfts/blob/4fb5720685626140b7f9982ea1e364faa6a27575/src/codegen/generate.ts#L231

rtfpessoa:
@fgnass do you have any suggestions on how to fix this?

kahirokunn:
Dot and hyphen symbols have a similar problem. 😢

pollen8:
thanks for fixing that - any chance we can get a release with these change?

tevariou:
> thanks for fixing that - any chance we can get a release with these change?

I concur. @fgnass a new release would be great!

fgnass:
@pollen8 If I'm not mistaken this was fixed by 25d67ba6519be8ad41661a15a412d5167b847c82 which was included in oazapfts@3.2.0. Please let me know if I am missing something and this is a different issue.

pollen8:
I'm not sure if this conforms to the swagger spec but now I have a proto file 

```
    /**
     * This is some comment in the proto
     */
    privacy_level?: PrivacyLevel;
}
```

which resolves to a swagger definition:

```
 "privacy_level": {
          "$ref": "#/definitions/PrivacyLevel",
          "title": "This is some comment in the proto"
        }
```
        And generates types:

 ```
        privacy_level?: ThisIsSomCommentInTheProto
 ```
        
With long descriptions this becomes pretty unwieldy

pollen8:
just to confirm that I re-tested this with 3.3.1 and can no longer replicate

PR #48, opened 2021-02-05
fix(generate): remove spaces from title

limawebdev1:
Addresses: https://github.com/cellular/oazapfts/issues/40

Hi there! Opened a PR to address this. Definitely could've done some regex replacement, but I referenced this stack overflow on speed and decided to go with plain ol' String/Array methods. https://stackoverflow.com/questions/5963182/how-to-remove-spaces-from-a-string-using-javascript

In OpenAPI spec, the `title` attribute is often used for a human-readable description of the schema, so definitely feel like there are many use cases where there'd be spaces in the title 😄. In my use case, I was uploading my OpenAPI spec to readme.io and it was nice to have spaces in the title for readability.

Totally open to feedback, thanks!

### DIFF
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -228,7 +228,7 @@ export default function generateApi(spec: OpenAPIV3.Document, opts?: Opts) {
     if (!ref) {
       const schema = resolve<OpenAPIV3.SchemaObject>(obj);
       const name = getUniqueAlias(
-        _.upperFirst(schema.title || getRefBasename($ref))
+        _.upperFirst(schema.title?.split(' ').join('') || getRefBasename($ref))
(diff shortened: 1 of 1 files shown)


stewartjarod:
This is a step in the right direction, but there are other known issues: https://github.com/cellular/oazapfts/issues/40#issuecomment-741749286

limawebdev1:
> This is a step in the right direction, but there are other known issues: [#40 (comment)](https://github.com/cellular/oazapfts/issues/40#issuecomment-741749286)

yep, saw that! if i have time later, i'll see if those are easy fixes 💜

fgnass:
Fixed in oazapfts@3.2.0

Issue #54, opened 2021-03-13
Missing `Blob` type for body in runtime fetchJson


tmueller:
Hi oazapfts team,

when using an upload endpoint, that returns JSON, code is generated, that does not compile. 

The Problem occurs, if the enpoints requestBodys schema has format `binary` and its response type is `application/json`. Then a function is generated, where the body parameter is expected to be `Blob`. This body parameter is passed on to `oazapfts.fetchJson`, where body is typed as `string | FormData` without support for `Blob`. 

A [recent change](https://github.com/cellular/oazapfts/blob/81513f814302c76a8254865eaf2e0b0aa0bb9b2d/src/codegen/generate.ts#L401) introduced the change to Blob when the requestBodys schema has format binary.

I think, the Problem could be solved by widening the type of the body parameter in fetchJson. The underlying implementation should be able to cope.

Thanks for creating such a nice package and making it open source.

tmueller:
With help of [the petstore example](https://github.com/cellular/oazapfts/blob/310117038b6a5119c8a121063d9f4d6f36f43b59/demo/petstore.json#L388) a similar scenario is tested. The behaviour I described does not occur during testing and all tests succeed. 

The reason is, that in the given example the requestBodies content is [multipart/form-data](https://github.com/cellular/oazapfts/blob/310117038b6a5119c8a121063d9f4d6f36f43b59/demo/petstore.json#L406). Oazapfts has a special handling for this case.

As of [swagger docs for file upload](https://swagger.io/docs/specification/describing-request-body/file-upload/), `multipart/form-data` might not be used. It is allowed to specify the requests media type, e.g. `image/png` for images of type png. This leads to the error I described above.

I'd be happy to submit a PR, I just need some more information:
Should I place an initially failing test in petstore.json?
Is widening the type of property `body` in [FetchRequestOpts](https://github.com/cellular/oazapfts/blob/25d67ba6519be8ad41661a15a412d5167b847c82/src/runtime/index.ts#L12) to include `Blob` considered the correct fix or are there any further undesired consequences?

fgnass:
Hi Thomas, a PR would be highly appreciated as I personally don't use oazapfts' blob support at all. Both adding a test to petstore.json as well as widening the type of body sound like a good approach to me!

PR #55, opened 2021-03-17
Extended blob support

tmueller:
Hi cellular team,

I started implementing Blob support as described in #54. Failing Tests as well as the widened type is implemented now. Currently I am struggeling to make the newly added test pass. 

I understand, [open-api-mocker](https://www.npmjs.com/package/open-api-mocker) is being used to mock an api based on petstore.json. All tests pass except my newly added test, that tests blob support. `open-api-mocker` responds with statuscode 400, although I am sure, the test should pass. Unfortunately documentation for `open-api-mocker` is nearly non-existent. Do you have a hint how to debug `open-api-mocker`'s response. Or do you have a different idea for testing?

### DIFF
--- a/demo/api.test.ts
+++ b/demo/api.test.ts
@@ -1,6 +1,9 @@
-import { ok, handle, okify, optimistic } from "oazapfts/lib/index";
 import * as api from "./api";
 import * as optimisticApi from "./optimisticApi";
+
+import { handle, ok, okify, optimistic } from "oazapfts/lib/index";
+
+import Blob from "cross-blob";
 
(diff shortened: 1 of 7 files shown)


jonkoops (approved):

jonkoops:
@tmueller I managed to get the test working. Could you verify that this implementation is working as expected for you?

tmueller:
@jonkoops Thanks for fixing the test case. Was [this commit](https://github.com/cellular/oazapfts/pull/55/commits/1d586c97aa0d57bb5e04e873dd897388f2a026a0) all it took? 🙈

I can verify, that this is indeed working very nicely. Tested in an express server.

PR #83, opened 2021-06-28
Do not convert bracket-style deepObject parameters for newer OpenAPI

jonkoops:
This prevents newer (v3) OpenAPI specs from being parsed with the legacy behavior for parameters using square brackets to represent a 'deepObject'. In OpenAPI v3 one can simply provide `style: deepObject` to represent these types, and provide a `type: object` with the correct fields.

For more information about these differences, see this [Stack Overflow post](https://stackoverflow.com/a/48498055/1056885).

The main motivation we have to make this change is to make it possible to pass in a schema that has both `foo` and `foo[field]`. If we were to keep the legacy behavior this would mean that we would get two parameters from `supportDeepObjects()` with the same name, which produces incorrect code as we get a type with the same field appearing twice.

For example, the following schema will produce incorrect code without this change in this PR:
```yaml
openapi: 3.0.0
info:
  title: Sample API
  version: 0.1.0
paths:
  /users:
    get:
      summary: Returns a list of users.
      parameters:
        - in: query
          name: page
          schema:
            type: string
            nullable: true
        - in: query
          name: page[size]
          schema:
            type: integer
        - in: query
          name: page[order]
          schema:
            type: string
            enum: [asc, desc]
      responses:
        "200": # status code
          description: A JSON array of user names
          content:
            application/json:
              schema:
                type: array
                items:
                  type: string
```
## Generated code before this PR
```ts
export function getUsers({ page, page }: {
  page?: string | null;
  // This is not valid, a duplicate declaration is not allowed.
  page?: {
      size?: number;
      order?: "asc" | "desc";
  };
} = {}, opts?: Oazapfts.RequestOpts) {
  ...
}
```

## Generated code after this PR
```ts
export function getUsers({ page, pageSize, pageOrder }: {
  page?: string | null;
  pageSize?: number;
  pageOrder?: "asc" | "desc";
} = {}, opts?: Oazapfts.RequestOpts) {
  ...
}
```

### DIFF
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -179,7 +179,9 @@ export function supportDeepObjects(params: OpenAPIV3.ParameterObject[]) {
 export default class ApiGenerator {
   constructor(
     public readonly spec: OpenAPIV3.Document,
-    public readonly opts: Opts = {}
+    public readonly opts: Opts = {},
+    /** Indicates if the document was converted from an older version of the OpenAPI specification. */
+    public readonly isConverted = false
   ) {}
 
   aliases: ts.TypeAliasDeclaration[] = [];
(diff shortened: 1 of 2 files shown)


jonkoops:
@fgnass I know you are not actively working for Cellular anymore, but since this is not a simple dependency upgrade I would love to have your input. Also if possible it would be great if we could get a release that includes this change.

fgnass:
LGTM. I can add you on npm too if you want.  Is your username the same as here?

jonkoops:
> LGTM. I can add you on npm too if you want. Is your username the same as here?

Awesome, thanks for the review! My NPM is indeed [jonkoops](https://www.npmjs.com/~jonkoops).

fgnass:
@jonkoops I've added you as maintainer on npm

PR #87, opened 2021-06-30
Return response headers

ramosbugs:
This PR exposes the response headers returned by the Fetch API,
which is helpful for extracting request IDs for error reporting, and
potentially for other use cases.


iggith:
Another use case is a `POST` request ending with creation (201) and returning an url of the newly created resource in the `Location` header ([RFC7231, sec. 6.3.2](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.2)). Without an access to headers the most important part of the result is missing with current implementation.

src/runtime/index.ts
```diff
@@ -56,10 +61,10 @@ export function runtime(defaults: RequestOpts) {
       : false;
 
     if (isJson) {
-      return { status, data: data ? JSON.parse(data) : null } as T;
+      return { status, data: data ? JSON.parse(data) : null, headers } as T;
     }
 
-    return { status, data } as T;
+    return { status, data, headers } as T;
```
jonkoops, line 67: Instead of picking the different fields here I think it might be better to just expose the original response entirely so that all of the fields can be accessed.

jonkoops (changes requested):

src/runtime/index.ts
ramosbugs: This required some non-trivial typing changes (e.g., to `ApiResponse`), but my project that depends on this library type checked without any changes using the latest revision.

fgnass:
@ramosbugs and @iggith: I created an alternative PR that addresses this issue by modifying the runtime and keeping the generated types untouched: https://github.com/oazapfts/oazapfts/pull/344. Let me know if this would work for you.

Issue #34, opened 2020-10-06
Incorrect ts code generated when using a ref again within anyOf, allOf, oneOf


anantoghosh:
Related to https://github.com/cellular/oazapfts/issues/31

If the same ref is used twice within `anyOf`, `allOf` or `oneOf`, incorrect ts code is generated.
```yaml
openapi: 3.0.3
info:
  title: PI
  version: 0.1.0
paths:
  /filter:
    get:
      description: stats
      parameters:
        - name: param1
          in: query
          schema:
            oneOf:
              - $ref: './datetime.yaml'
              - $ref: './rating.yaml'
        - name: param2
          in: query
          schema:
            oneOf:
              - $ref: './datetime.yaml'
              - $ref: './rating.yaml'
      responses:
        "200":
          description: ok
```

datetime.yaml
```yaml
type: string
```
rating.yaml
```yaml
type: integer
```

Generates
```js
export type 0 = string; // ERROR
export type 1 = number; // ERROR
/**
 * Agent stats
 */
export function getFilter({ param1, param2 }: {
    param1?: string | number;
    param2?: 0 | 1;
} = {}, opts?: Oazapfts.RequestOpts) {
    return oazapfts.ok(oazapfts.fetchText(`/filter${QS.query(QS.form({
        param1,
        param2
    }))}`, {
        ...opts
    }));
}
```

## Here the ref paths were generated as
```js
"schema": {
  "oneOf": [
    {
      "$ref": "#/paths/~1filter/get/parameters/0/schema/oneOf/0"
    },
    {
      "$ref": "#/paths/~1filter/get/parameters/0/schema/oneOf/1"
    }
  ]
}
```

Morriz:
Any updates here? Does this mean we can't use refs inside anyOf/allOf/oneOf? That would be painful.

fgnass:
Closed by 1aee9f9

PR #125, opened 2021-08-26
Avoid generating type identifiers starting with numbers

vpzomtrrfrt:
Not sure if this is the right method to change, and there are probably still other cases where invalid identifiers are generated, but this resolves the specific case of refs with numbers.

Somewhat addresses #34 though the code is still weird


fgnass:
Fixed by 90c2610

PR #143, opened 2021-09-20
feat: query string parameter content support

sstone1:
We use LoopBack (https://loopback.io/), which generates APIs that have complex query string parameters which must be serialized as JSON before adding them to the URL. These are specified in OpenAPI as query string parameters with a `content` property, similar to the following:

```json
{
  "in": "query",
  "name": "filter",
  "content": {
    "application/json": {
      "schema": {
        "properties": {
          "where": {
            "properties": {
              "fileId": {
                "type": "integer"
              }
            },
            "type": "object"
          }
        },
        "type": "object"
      }
    }
  }
}
```

Currently the code generated tries to encode these as form parameters, which fails.

These changes update the code generator to handle these parameters correctly, and serialize them as JSON.

Signed-off-by: Simon Stone <Simon.Stone@docusign.com>

### DIFF
--- a/demo/contentParams.json
+++ b/demo/contentParams.json
@@ -1,0 +1,56 @@
+{
+  "info": {
+    "title": "example with parameters specified with content",
+    "version": "0.0.1"
+  },
+  "openapi": "3.0.0",
+  "servers": [
+    {
+      "url": "https://api.example.com"
+    }
+  ],
+  "paths": {
+    "/file": {
+      "get": {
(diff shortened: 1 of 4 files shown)


shocklateboy92 (approved):
Okay, I looked at the [OpenAPI spec](https://swagger.io/docs/specification/describing-parameters/#schema-vs-content) and this does make sense.
Could you please rebase on latest master?

fgnass:
Merged in #346

PR #120, opened 2021-08-21
Add support for properties outside allOf

piotr121993:

PR #207, opened 2022-02-05
Account for properties and additionalProperties in allOf

msutkowski:
Hi there! We've had a handful of issues reporting inaccurate type generation for users of [RTK Query Code Generator](https://github.com/reduxjs/redux-toolkit/tree/master/packages/rtk-query-codegen-openapi) which leverages oazapfts.

I made a quick reproduction comparing oazapfts output to a user-provided schema to various other generators here: https://github.com/msutkowski/oazapfts-typesecript-codegen-repro. You can see that other generators will produce a type like:

`type Dog = Animal & { uniqueDogProp?: string | null }`. 

With the current version of oazapfts, we were generating

`type Dog = Animal`

This PR addresses that by accounting for `properties` and `additionalProperties` when handling the `schema.allOf` case. The diff in the output looks like this against the repro above:

image

Referenced issues with reports:

- https://github.com/reduxjs/redux-toolkit/issues/2002
- https://github.com/reduxjs/redux-toolkit/issues/1940

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1222,9 +1222,9 @@
       }
     },
     "node_modules/@types/lodash": {
-      "version": "4.14.186",
-      "resolved": "https://registry.npmjs.org/@types/lodash/-/lodash-4.14.186.tgz",
-      "integrity": "sha512-eHcVlLXP0c2FlMPm56ITode2AgLMSa6aJ05JTTbYbI+7EMkCEE5qk2E41d5g2lCVTqRe0GnnRFurmlCsDODrPw==",
+      "version": "4.14.188",
+      "resolved": "https://registry.npmjs.org/@types/lodash/-/lodash-4.14.188.tgz",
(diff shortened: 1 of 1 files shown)


msutkowski:
Ah, should've checked before I opened this - but it's a duplicate of #120

msutkowski:
Just noting for anyone else that might be following this on RTKQ's behalf - we're going to start [maintaining a patched version](https://github.com/rtk-incubator/oazapfts) of this for issues that impact our users directly.

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
jonkoops, line 1: Could we remove the Prettier config from this PR? Seems out of scope for the changes here.

jonkoops (changes requested):
Other changes look good. If we can get rid of the Prettier config I think we can merge this.

.prettierrc
msutkowski: removed.

Xiphe:
oh, were the changes here also part of https://github.com/cellular/oazapfts/pull/234 ?

PR #235, opened 2022-05-31
Add .prettierrc

kahirokunn:


kahirokunn:
@jonkoops https://github.com/cellular/oazapfts/pull/234#discussion_r885580897
How about now?

jonkoops:
I think if we were to go for a linter/formatter I think I'd like to see a couple of things:

1. Only using the defaults, I like to have an opinionated style guide without the need to keep to much config around.
2. It should be enforced by both the CI as well as when commits are made (i.e. `lint-staged`).

kahirokunn:
@jonkoops 
Okay! How about now?

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
jonkoops, line 1: Let's use the default Prettier settings, I like to keep things opinionated.

package.json
```diff
@@ -8,6 +8,8 @@
   },
   "scripts": {
     "build": "tsc",
+    "lint": "npx prettier '**/*.{js,jsx,ts,tsx,json}' --check",
+    "lint:fix": "npx prettier '**/*.{js,jsx,ts,tsx,json}' --write",
```
jonkoops, line 12: Since `prettier` is already installed as a dev dependency there is no need to use `npx`. I also believe that we can omit the extensions and do the whole directory with all of the Prettier supported extensions instead. I'd also like `--write` to be the default so we'd end up with something like this.

```suggestion
    "lint": "prettier . --write",
```

package.json
```diff
@@ -29,9 +31,9 @@
     "@types/minimist": "^1.2.0",
     "@types/node": "^17.0.22",
     "form-data": "^4.0.0",
-    "husky": "^7.0.0",
+    "husky": "^8.0.1",
     "jest": "^27.0.5",
-    "lint-staged": "^12.0.2",
+    "lint-staged": "^12.5.0",
```
jonkoops, line 36: There is no need to configure `lint-staged`? It will just detect the Prettier config?

jonkoops (changes requested):

Xiphe:
Isn't prettier already installed? + This seems stale, closing.

PR #238, opened 2022-06-12
cleanup duplicated code in keywordType, etc.

### LINKED ISSUE #237: Repetitive code between `function createKeywordType` and `const keywordType`
I was looking at the source code and I noticed that the same mapping of keyword name to TS node is happening in two places in the `tscodegen.ts` file. `export function createKeywordType` and `export const keywordType` have the same logic and can definitely be combined.  

It seems as though the `function createKeywordType` is not actually used anywhere.  Can it be deleted?  Or is it part of the public API?

https://github.com/cellular/oazapfts/blob/0f6225941973bada228b27d1fb60a09462d40a60/src/codegen/tscodegen.ts#L22-L41

https://github.com/cellular/oazapfts/blob/310117038b6a5119c8a121063d9f4d6f36f43b59/src/codegen/tscodegen.ts#L43-L55

I haven't dug through the entire codebase but I am inclined to think that the explicit type annotation on `const keywordType` is more harmful than helpful, as it:

- Allows any `string` key, not just the valid keys.
- Prevents intellisense suggestions.
- Broadens the type of each property to the union rather than its specific type.

 There are actually a few places in the code that use `cg.keywordType.void` even though it doesn't exist!  This would be an error without the explicit type on `keywordType`.

lindapaiste:
Fixes #237 

`keywordType`
- Removed the TypeScript type on `const keywordType` so that a more specific type can be inferred.
- Add `keywordType.void` which was previously used but never defined.
- Refactor `function createKeywordType` to use `keywordType` internally.
- Derive the `type KeywordTypeName` from `keyof typeof keywordType` instead of typing out all the strings.
- Swapped lines 415 and 416 in `generate.ts` so that the second line gets a narrowed type.  This avoids TS errors in 4.6 but might be an error in lower 4.x versions, so I'm not sure if I should put an `as` assertion there?

`createEnumTypeNode`
- Combine enum creating logic from `generateServers.ts` and `generate.ts`.
- Move local function `createLiteral` from `generateServers.ts` into `tscodegen.ts` so that it can be reused.
- Create new utility `createEnumTypeNode` in `tscodegen.ts` which replaces the block of code in `generate.ts` 394-408 and the local function `createUnion` in `generateServers.ts`.

Misc
- `generateServers.ts`
  - Use existing utility function `cg.createTemplateString` inside of `createTemplate`.
  - Move the `props` variable (line 103) inside of the function call to eliminate the `as` assertion.
- `generate.ts`
  - Added some return types on methods and replaced a few `: any` with specific types.
- `tscodegen.ts`
  - Removed the `ts.parseIsolatedEntityName;` at the top of the file which I think is a mistake?  It was added [here](https://github.com/cellular/oazapfts/commit/c3c33f8ed285ff6f3e84b375785417dc06b141df).

~~It looks like I made way more changes because of prettier 🤷‍♀️ Maybe I should make an additional PR that just applies prettier to `src/codegen/generate.ts` and have that be merged first?~~

### DIFF
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -1,41 +1,41 @@
-import _ from "lodash";
-import ts, { factory } from "typescript";
-import path from "path";
-import { OpenAPIV3 } from "openapi-types";
-import * as cg from "./tscodegen";
-import generateServers, { defaultBaseUrl } from "./generateServers";
-import { Opts } from ".";
-import { threadId } from "worker_threads";
+import _ from 'lodash';
+import ts, { factory } from 'typescript';
+import path from 'path';
+import { OpenAPIV3 } from 'openapi-types';
+import * as cg from './tscodegen';
+import generateServers, { defaultBaseUrl } from './generateServers';
+import { Opts } from '.';
+import { threadId } from 'worker_threads';
 
 export const verbs = [
-  "GET",
-  "PUT",
-  "POST",
-  "DELETE",
-  "OPTIONS",
-  "HEAD",
-  "PATCH",
-  "TRACE",
+  'GET',
+  'PUT',
+  'POST',
+  'DELETE',
+  'OPTIONS',
+  'HEAD',
+  'PATCH',
+  'TRACE',
 ];
 
-type ContentType = "json" | "form" | "multipart";
+type ContentType = 'json' | 'form' | 'multipart';
 
 export const contentTypes: Record<string, ContentType> = {
-  "*/*": "json",
-  "application/json": "json",
-  "application/hal+json": "json",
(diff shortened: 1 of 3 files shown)


lindapaiste:
> Can we get rid of the code style changes in this PR?

I'll see what I can do.  I think I have to temporarily take the pre-commit script out of package.json, otherwise the prettier rewrites will happen automatically on every commit.
```
  "lint-staged": {
    "*.{ts,js,json,md,yml}": "prettier --write"
  }
```

Xiphe:
LGTM :shell:

fgnass:
Merged in #346

PR #242, opened 2022-06-30
Use built-in Fetch API from Node.js

jonkoops:


Xiphe:
worth mentioning that this is only about tests and dev dependencies. The library never had node-fetch as a dependency and required passing a fetch implementation upon usage.

LGTM :hamster:

Xiphe (approved):

Xiphe:
Tried to debug why ts-jest refuses to use build in fetch from node and gave up for now.

jonkoops:
@Xiphe I have managed to get this working, and was able to remove both `node-fetch` and `cross-blob`. However, I am seeing some test failures that are seemingly unrelated.

It looks like OpenAPI Mocker returns invalid responses for some of the POST calls. It seems like it's not returning the updated responses.

Any idea what might be causing this?

Xiphe:
What I found out: When you run these tests in isolation on `master` they fail in the same way. 

I don't have any more time to dig deeper but this indicates race condition to me :)

jonkoops:
Yeah, I would concur. Very peculiar. Perhaps this is a bug in OpenAPI mocker?

Xiphe:
Seems as if the same error happens also at random: https://github.com/cellular/oazapfts/pull/311 https://github.com/cellular/oazapfts/actions/runs/3310145388/jobs/5464028024

Xiphe:
seems as if the current state on `master` is flaky

jonkoops:
Ah, that's not good. I'll see if I can take a look at that. Perhaps we just skip verifying the exact contents of the response and just check that the response is actually ok?

Xiphe:
might be a sensible solution. I'll also try to have a look and debug this week

jonkoops:
Updated the tests so the `status` field is no longer present. We still might want to see if we can find a fix, but for now this allows the tests to pass.

PR #258, opened 2022-07-18
Allow application/geo+json response types

olof-nord:
Some well defined apis define their geojson types in OpenAPI specs, and those may be of the type `application/geo+json`: https://datatracker.ietf.org/doc/html/rfc7946. 

This include some of the API specs I am using, which when using this type today to render TS oazapfts unfortunately ends up using the Blob data type.

This PR adds support for this content type just as it has for `application/hal+json` and `application/problem+json`.

In short, this PR is based on the recently merged https://github.com/cellular/oazapfts/pull/197 one.

### DIFF
--- a/demo/geojson.json
+++ b/demo/geojson.json
@@ -1,0 +1,83 @@
+{
+  "info": {
+    "title": "example with GeoJSON download",
+    "version": "0.0.1",
+    "description": "The example is taken from RFC 7946, 1.5: Example"
+  },
+  "openapi": "3.0.0",
+  "servers": [
+    {
+      "url": "https://api.example.com"
+    }
+  ],
+  "paths": {
+    "/geojson": {
+      "get": {
+        "operationId": "GetGeoJSON",
+        "description": "Get GeoJSON JSON data",
+        "responses": {
+          "200": {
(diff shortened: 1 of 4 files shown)


shocklateboy92 (changes requested):
Hi @olof-nord and thanks for your PR!

The change seems sensible, but could you please add a test that covers the new behavior.

jonkoops:
I'm starting to think that we should perhaps change this to some kind of regular expression that checks the mime type for `application/*+json`, or perhaps provide an option to pass a function that allows the mime-type to be checked in user-space.

shocklateboy92 (approved):
I'm actually okay with this. We can do a regex if the cases get too unwieldy.
Also happy this increases our test coverage (I think) :test_tube: 

However, I'll give a few days before merging so @jonkoops can chime in if he feels strongly about the regex.

jonkoops (approved):
LGTM

jonkoops:
I think extending this for now is more than fine, but might get unwieldy as we go.

olof-nord:
Thanks for the good feedback, always happy to increase that test coverage 👍

PR #280, opened 2022-08-25
serialize object query parameters as exploded form by default

Xiphe:
As described in https://github.com/cellular/oazapfts/issues/269, we were not handling object parameters according to swagger specification.

BREAKING CHANGE:
When `style` and/or `explode` are not explicitly set on a path parameter in your swagger.json
Then in the past oazapfts had send an array as `?key=a,b`
From now on the same data is sent as `?key=a&key=b`
Also space and pipe delimited serialization now requires explicit disabling of explode to be effective

**Migration:**
When your backend implements the OpenAPI Specification correctly, you do not need to do anything.

If not, or you're unsure, in your swagger.json:
- Set `"explode": false` on all array and object path parameters that had no explicit explode setting before
- Except for object parameters with `"style": "deepObject"`. These have been handled correctly and you should not change them.

See https://swagger.io/docs/specification/serialization/#query for details

---

Props to @jannik-hh who discovered the issue

### DIFF
--- a/demo/api.test.ts
+++ b/demo/api.test.ts
@@ -7,6 +7,7 @@ api.defaults.baseUrl = `${process.env.SERVER_URL}/v2`;
 optimisticApi.defaults.baseUrl = `${process.env.SERVER_URL}/v2`;
 
 (global as any).fetch = require("node-fetch");
+(global as any).Response = require("node-fetch").Response;
 (global as any).FormData = require("form-data");
 
 describe("ok", () => {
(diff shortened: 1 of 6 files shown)


Xiphe:
PTAL @jonkoops || @fgnass

shocklateboy92 (approved):
LGTM. It's a good idea to make oazapfts more compliant with OpenAPI spec.
Unless @fgnass or @jonkoops has any objections, I'll merge this in a few days.

fgnass (approved):

Xiphe:
Published as `v4.0.0`

PR #292, opened 2022-09-14
feat: add support for 'const' keyword

### LINKED ISSUE #291: feat: support "const" keyword
Would you accept a PR for adding support for the "const" keyword? It's basically syntactic sugar for a single value enum, however its becoming more popular in the wild since it's been a part of the JSON schema spec since draft-06 and is now a part of OAS 3.1. 

I know the current implementation brings in types from OAS 3.0 so is this out of the question until the library makes a decision to support 3.1?

smiley-uriux:
Closes #291 

I implemented this by strictly interpreting the spec which states that "const" is just shorthand for an enum with a single value.

I did notice that enum values right now are restricted to primitive types, when technically object literals are valid (though I've never had a use case for them). I started to take a stab at providing support, but it was a little trickier than I thought since an ObjectLiteralExpression does not extend TypeNode, and I couldn't figure out how to reconcile that difference without altering the fundamental return type of many of the library's methods. Regardless, it was sufficiently complex enough to deserve its own PR if there is interest.

### DIFF
--- a/src/codegen/generate.test.ts
+++ b/src/codegen/generate.test.ts
@@ -55,6 +55,15 @@ describe("generate", () => {
 
     expect(oneLine).toContain(circleTypeDefinition);
   });
+
+  it("should handle enums", async () => {
+    const api = printAst(new ApiGenerator(spec.petstore).generateApi());
+    const oneLine = api.replace(/\s+/g, " ");
+
+    const enumTypeDefinition = `export type Option = ("one" | "two" | "three")[];`;
+
+    expect(oneLine).toContain(enumTypeDefinition);
+  });
 });
 
 describe("generate with application/geo+json", () => {
@@ -90,5 +99,22 @@ describe("generate with blob download", () => {
     expect(oneLine).toContain(
       "return oazapfts.fetchBlob<{ status: 200; data: Blob; }>(`/file/${encodeURIComponent(fileId)}/download`, { ...opts });"
     );
+  });
+});
+
+describe("generate with const", () => {
+  let spec: OpenAPIV3.Document;
+
+  beforeAll(async () => {
+    spec = (await SwaggerParser.bundle(
+      __dirname + "/../../demo/const.json"
+    )) as any;
+  });
+
+  it("should generate an api with literal type set to const value", async () => {
+    const artefact = printAst(new ApiGenerator(spec).generateApi());
+    const oneLine = artefact.replace(/\s+/g, " ");
+    const constTypeDefinition = `export type Shape = "circle";`;
+    expect(oneLine).toContain(constTypeDefinition);
   });
 });
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -5,7 +5,6 @@ import { OpenAPIV3 } from "openapi-types";
 import * as cg from "./tscodegen";
 import generateServers, { defaultBaseUrl } from "./generateServers";
 import { Opts } from ".";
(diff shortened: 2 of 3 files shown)


src/codegen/generate.ts
```diff
@@ -30,6 +30,11 @@ export const contentTypes: Record<string, ContentType> = {
   "multipart/form-data": "multipart",
 };
 
+// augment SchemaObject type to allow slowly adopting new OAS3.1+ features
+type SchemaObject = OpenAPIV3.SchemaObject & {
+  const?: any;
```
shocklateboy92, line 35: Not a big fan of `any`. I would very much prefer an `unknown` instead.
Should be an easy change since you're already doing `typeof` checks where it's used.

shocklateboy92 (changes requested):
Thanks for your PR - it's a neat feature. :bow: 
Please add tests to cover your new case. :test_tube:

src/codegen/generate.ts
smiley-uriux: The decision to use "any" was because the `enum` prop is currently typed as "any[]", and since this is meant to be a shorthand for a single-value enum I didn't want to introduce any potential type discrepancies. Either way will compile fine and neither provide much protection in the current code, so I'm happy to change it.

src/codegen/generate.test.ts
```diff
@@ -55,6 +55,15 @@ describe("generate", () => {
...
   });
+
+  it("should handle enums", async () => {
+    const api = printAst(new ApiGenerator(spec.petstore).generateApi());
+    const oneLine = api.replace(/\s+/g, " ");
+
+    const enumTypeDefinition = `export type Option = ("one" | "two" | "three")[];`;
+
+    expect(oneLine).toContain(enumTypeDefinition);
+  });
```
Xiphe, line 66: This tests as-is behavior, right?

Xiphe (approved):
LGTM :hatched_chick:

Pre-released this as `v4.1.0-const-support.1` try it out via `npm i oazapfts@const-support`

src/codegen/generate.test.ts
smiley-uriux: Yes exactly. I moved around some of the enum processing code so wanted to have a little test coverage there to make sure I didn't break existing behavior. I would have used the existing pet.json example for the "const" test, but since "const" is technically OAS 3.1, I didn't want to add anything out of spec to that file. If more 3.1 features make it in, it will probably make sense to keep a pet-3.1.json example file around that future tests can go off of.

Xiphe: Cool! Thanks for elaborating and for taking the extra afford 👍

src/codegen/generate.ts
```diff
@@ -444,18 +437,36 @@ export default class ApiGenerator {
     return cg.keywordType.any;
   }
 
+  /**
+   * Creates literal type (or union) from an array of values
+   */
+  getTypeFromEnum(values: any[]) {
```
shocklateboy92, line 443: Very minor point, but if you could make this `any[]` an `unknown[]` as well, I think that would be better.
(I understand your point about `schema.enum` being an `any[]`, but I rather like the `unknown` type :grin: ).

shocklateboy92 (approved):

src/codegen/generate.ts
Xiphe: will add this and release 🚀

Xiphe:
released as `oazapfts@4.1.0`

PR #295, opened 2022-09-18
fix: remove duplicate parameters

### LINKED ISSUE #58: duplicate argument
using https://raw.githubusercontent.com/prometheus/alertmanager/2b6315f399b95c4e327dcd5f5eb10866ef844252/api/v2/openapi.yaml

result:
```
export function deleteSilence(silenceId: string, silenceId: string, opts?: Oazapfts.RequestOpts) {
```

jordanbtucker:
When OpenAPI specs reference parameters more than once in the same hierarchy, like this snippet from Asana's OpenAPI spec, oazapfts produces duplicate parameters. This fixes that by removing duplicate parameters from `resolvedParameters`.

```yaml
  /projects/{project_gid}/project_statuses:
    parameters:
      - $ref: '#/components/parameters/project_path_gid'
      - $ref: '#/components/parameters/pretty'
      - $ref: '#/components/parameters/fields'
    get:
      summary: Get statuses from a project
      description: |-
        *Deprecated: new integrations should prefer the `/status_updates` route.*

        Returns the compact project status update records for all updates on the project.
      tags:
        - Project Statuses
      operationId: getProjectStatusesForProject
      parameters:
        - $ref: '#/components/parameters/project_path_gid'
        - $ref: '#/components/parameters/pretty'
        - $ref: '#/components/parameters/fields'
        - $ref: '#/components/parameters/limit'
        - $ref: '#/components/parameters/offset'
```

Fixes #58

### DIFF
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -673,10 +673,10 @@ export default class ApiGenerator {
         }
 
         // merge item and op parameters
-        const resolvedParameters = [
+        const resolvedParameters = _.uniq([
           ...this.resolveArray(item.parameters),
           ...this.resolveArray(op.parameters),
-        ];
+        ]);
 
         // expand older OpenAPI parameters into deepObject style where needed
         const parameters = this.isConverted


shocklateboy92 (changes requested):
Could you please add a test that covers this case?
So that way, future code changes don't re-introduce this bug :racing_car:

Xiphe:
Thanks for the contribution @jordanbtucker!
I agree with @shocklateboy92, a test for the bug would be great! Let us know when you need any help with that or wan't me to add the test.

fgnass:
This is fixed by 3d4634cdfddbd9b7526d30143c5b2d2354eaf9a1 which does not only check for object identity but drops params that have the same `name` and  `in` (favouring the param defined deeper down the hierarchy).

PR #313, opened 2022-10-26
fix pronunciation YouTube link

aSemy:
When I click on the pronunciation link I'm directed to go to https://youtu.be/chvb-K95rBE, but that page doesn't load for me.

image

Maybe it's something that's wrong with my network or computer, but the full link works for me https://www.youtube.com/watch?v=chvb-K95rBE


Xiphe:
The minified link works on my end. Still i don't mind the change...

PR #330, opened 2022-11-18
Fix body type in JsonRequestOpts to allow string (and any other type supported by JSON.stringify)

swernerx:
We just started using this library. When sending over the library to TypeScript for building a production-ready library we see errors related to the `body` property of the `JsonRequestOpts` type defined in the runtime. The value of `body` is passed to `JSON.stringify` to send over to the server next. So all types supported by this method should be valid here. As `JSON.stringify` defines this first parameter to be `any`, I used the same here for `JsonRequestOpts`.

I verified that it works now by executing:

- npm run test:unit
- npm run build

Both worked without issues.

I also updated a few dependencies in here. Mostly very minor, but bumped TypeScript to v4.7. Later versions seem to have a breaking change and require some more modifications. Interestingly, the previous version lock file was locked to 4.8.4... I am not actually sure whether this is an issue.


swernerx:
Bildschirmfoto 2022-11-18 um 13 30 24

That's the error we got.

Xiphe:
Hey, thanks for taking the time to contribute. Could you help me understand the issue better by either sharing the part of the schema that causes this error or even better provide a test that replicates the issue?

swernerx:
I'm not very deep in this area of OpenAPI. 

This is one of the generated outputs from our schema files:

```code
/**
 * Change the state of an existing offer.
 */
export function offersSetState(tenant: string, id: string, ioEnpowerxResourcesV1OfferState: IoEnpowerxResourcesV1OfferState, opts?: Oazapfts.RequestOpts) {
    return oazapfts.fetchJson<{
        status: 200;
        data: IoEnpowerxResourcesV1Offer;
    } | {
        status: 400;
        data: any;
    } | {
        status: 401;
        data: any;
    } | {
        status: 403;
        data: any;
    } | {
        status: 404;
        data: any;
    } | {
        status: 409;
        data: any;
    } | {
        status: 429;
        data: any;
    } | {
        status: 500;
        data: any;
    } | {
        status: 501;
        data: any;
    } | {
        status: 503;
        data: any;
    } | {
        status: 400.0;
        data: any;
    } | {
        status: 400.1;
        data: any;
    } | {
        status: 409.2;
        data: any;
    }>(`/resources/v1/tenants/${encodeURIComponent(tenant)}/offers/${encodeURIComponent(id)}/state`, oazapfts.json({
        ...opts,
        method: "PUT",
        body: ioEnpowerxResourcesV1OfferState
    }));
}
```

This is the section from the schema file:

```code
"parameters": [
          {
            "name": "tenant",
            "description": "The tenant the offer belongs to.",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "id",
            "description": "The ID identifying the offer in the tenant's scope.",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "state",
            "description": "The state to set the offer to.",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/io.enpowerx.resources.v1.Offer.State"
            }
          }
        ],
```

The type is defined as:

```code
    "io.enpowerx.resources.v1.Offer.State": {
      "type": "string",
      "enum": [
        "STATE_UNSPECIFIED",
        "PREPARED",
        "OFFERED",
        "ACCEPTED",
        "DECLINED",
        "EXPIRED"
      ],
      "default": "STATE_UNSPECIFIED",
      "description": " - STATE_UNSPECIFIED: The offer state is unspecified.\n - PREPARED: The offer is ready but has not yet been sent to the customer.\n - OFFERED: The offer has been sent to the customer and is awaiting a response.\n - ACCEPTED: The customer accepted the offer.\n - DECLINED: The offer declined the offer.\n - EXPIRED: The offer has expired."
    },
```

That said... I wonder why `object` was used in the first place. Generally that's a pretty weak type anyway... and effectively we don't really have to limit it there as long as `JSON.stringify` can transform it into a string.

Xiphe:
Thanks for the feedback! I agree, object seems not optimal here. Will look into this next week.

swernerx:
I pushed a small update to fix merge conflicts.

Xiphe:
Sorry that I have not managed to look into this, having a busy week. But it's on my list.

Xiphe:
Sorry for the late response. 

I'm having troubles re-creating the issue based on only the `parameters` of your method. Could you provide the full method spec including "requestBody"? So that I can confirm that this is an actual issue within this library and not caused by a non-standard configuration creating artifacts.

Speaking of non-standard:
From what I see the `"in": "body"` is not part of the [specification](https://swagger.io/specification/#parameter-object) and I'm unsure how well oazapfts takes this.

swernerx:
As said, I am not an expert... or even knowledgeable regarding swagger. The swagger files we use are generated by a tool from a protobuf/GRPC implementation. I guess I pass your feedback to some more knowledgeable people in my company.

Other than that, I figure that the TS restriction to `object` still might not make too much sense. To change that to `any` and following the `JSON.stringify` specification should not hurt and aligns it with the API being used. Also, this does not change support for any of the existing tests.

BTW: Interestingly, some sources assume that `"in": "body"` is okay: See this screenshot from: https://editor.swagger.io/:
Bildschirm­foto 2022-11-29 um 21 38 40

I am not sure where this is defined... maybe something old/deprecated?

swernerx:
Our swagger files are hosted online: https://apis.enpowerx.io/swagger/ - I used all of these files individually and passed them through [oazapfts](https://github.com/cellular/oazapfts)

swernerx:
I created a small swagger reduction of the issue:

Input:

```json
{
  "swagger": "2.0",
  "info": {
    "title": "EnPowerX Contracting API",
    "version": "V2 Beta",
    "contact": {
      "name": "EnPowerX APIs",
      "url": "https://apis.enpowerx.io"
    }
  },
  "schemes": [
    "https",
    "http"
  ],
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
  ],
  "paths": {
    "/tenants/{tenant}/contracting/v2/customers/{customer}/billing/contact/email": {
      "put": {
        "summary": "Change the email for all contracts of a specific customer.",
        "operationId": "ContractBilling_SetAllEmailAddresses",
        "responses": {
          "200": {
            "description": "A successful response.",
            "schema": {
              "type": "object",
              "properties": {}
            }
          }
        },
        "parameters": [
          {
            "name": "tenant",
            "description": "The customer's tenant.",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "customer",
            "description": "The customer ID.",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "email",
            "description": "The desired email address.",
            "in": "body",
            "required": true,
            "schema": {
              "type": "string",
              "required": [
                "email"
              ]
            }
          }
        ]
      }
    }
  }
}
```

Output:
```ts
export function contractBillingSetAllEmailAddresses(tenant: string, customer: string, body: string, opts?: Oazapfts.RequestOpts) {
    return oazapfts.fetchJson<{
        status: 200;
        data: {};
    }>(`/tenants/${encodeURIComponent(tenant)}/contracting/v2/customers/${encodeURIComponent(customer)}/billing/contact/email`, oazapfts.json({
        ...opts,
        method: "PUT",
        body
    }));
}
```

Xiphe:
You're right [v2.0 of the specification](https://swagger.io/specification/v2/) uses this notation for body parameters.

Xiphe:
Released as [`v4.3.1`](https://github.com/cellular/oazapfts/releases/tag/v4.3.1) thanks for the contribution 🎉

Xiphe:
Just realized `JsonRequestOpts` was also used for `form` serialization which suffers from the same issue but cant be fixed with only adjusting the type

swernerx:
I'm not sure what you mean right now? The generator runs successfully with v4.3.1 in any case. Is this another limitation in the current oazapfts, which currently does not affect us?

Xiphe:
No, this does not affect you as your body is JSON-serialized. Its just that the change you made initially also affected the parameters of the form serialization which would have failed to serialize a string so I've change the parameter type for that one.

Nothing to do for you, just wanted to add this here since the PR is referenced from the changeset.

PR #331, opened 2022-11-18
Format Path Parameters

Xiphe:
This aims to provide a non-breaking fix for https://github.com/cellular/oazapfts/issues/319 

But while working at it I realized that all the parameter formatting currently aims at query-strings and implicitly used defaults for them etc. 

I currently tend to not actually merge this but instead rework the parameter serialization again (flashback to https://github.com/cellular/oazapfts/pull/280) to be spec compliant for query, path, cookie and header parameters.

### DIFF
--- a/demo/api.ts
+++ b/demo/api.ts
@@ -542,10 +542,10 @@ export function getObjectParameters(
         commaArray,
         commaObject,
       }),
-      QS.space({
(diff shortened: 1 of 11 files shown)


Xiphe:
released as `4.3.1-formatted-path-params.1` so we can try this out.

PR #344, opened 2022-12-01
feat: expose response headers

fgnass:
This PR has the same intention as #87 but does not change any of the generated types, only the runtime.

I thought about @jonkoops' [comment](https://github.com/oazapfts/oazapfts/pull/87#discussion_r910902783) about exposing the whole request instead of just the headers, but didn't like the fact that one could get hold of the body this way. I also prefer the DX of writing `res.headers` rather than `res.response.headers`.

### DIFF
--- a/demo/api.test.ts
+++ b/demo/api.test.ts
@@ -327,4 +327,13 @@ describe("Blob", () => {
       message: "OK",
     });
   });
+
+  describe("headers", () => {
+    it("should return headers", async () => {
+      const res = await api.getPetById(1);
+      expect(res.headers.get("x-powered-by")).toBe(
+        "jormaechea/open-api-mocker"
+      );
+    });
+  });
 });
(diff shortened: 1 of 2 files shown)


iggith:
> (@fgnass on #87) @ramosbugs and @iggith: I created an alternative PR that addresses this issue by modifying the runtime and keeping the generated types untouched: https://github.com/oazapfts/oazapfts/pull/344. Let me know if this would work for you.

Yes. Having an access to the headers makes handling [RFC7231, sec. 6.3.2](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.2) with use of the `Location` header straightforward.

To say the truth in the mean time we decided to refactor the API to avoid headers and be able to use an official release of `oazapfts`. But anyway this was rather a tradeoff. Is good to see it will not be required in the future. Thank you for your effort!

Xiphe (approved):
LGTM :racehorse:

fgnass:
Closing in favour of #346

Issue #46, opened 2021-01-14
Feature request: Add support for application/ld+json


Richard87:
Hi!

I would love to have support for application/ld+json, and prefer it when generating types.
(https://github.com/rtk-incubator/rtk-query-codegen/issues/28) 

Would you accept a PR for that, or is it out of scope? :) 

And if so, have any pointers to where I should start? :)

fgnass:
It should be as simple as adding it to https://github.com/cellular/oazapfts/blob/main/src/codegen/generate.ts#L23.

That being said, it might also be a good idea to map every content-type that contains `json` to `json` without having to add it explicitly.

fgnass:
Closed by #346

PR #345, opened 2022-12-01
feat: support unknown json types

fgnass:
This PR addresses #46 (and other similar issues we had in the past). Instead of explicitly listing all kinds of application/json flavours, it checks if the content type contains the `json` token.

### DIFF
--- a/src/codegen/generate.test.ts
+++ b/src/codegen/generate.test.ts
@@ -1,4 +1,8 @@
-import ApiGenerator, { getOperationName } from "./generate";
+import ApiGenerator, {
+  getOperationName,
+  isJsonMimeType,
+  isMimeType,
+} from "./generate";
 import { printAst } from "./index";
 import SwaggerParser from "@apidevtools/swagger-parser";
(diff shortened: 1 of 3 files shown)


Xiphe (approved):
LGTM :baby_chick:

fgnass:
Closing this in favour of #346

fgnass:
Turns out is wasn't working as expected. It indeed was generating TypeScript code that did not even compile. The tests did not catch the error as we are only compiling the `petstore.json` API in `/demo` but not all the other fixtures we use in the unit tests. 🤦‍♂️

fgnass:
I'll try to add this to all the unit tests: https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#a-simple-transform-function ...this way we could ensure that they all produce valid code.

fgnass:
... and done. From now on all the code generated by our unit tests is type-checked: 4ed01a62f5a06cc8f10ba4dfdeed82e76dc235af

Issue #365, opened 2023-01-16
generated enum value is not correct


trim21:
json schema and generated code doesn't match

https://github.com/oazapfts/oazapfts/blob/b24c627e942580b2df56421739da9d438a0fa5f8/demo/petstore.json#L1223

https://github.com/oazapfts/oazapfts/blob/b24c627e942580b2df56421739da9d438a0fa5f8/demo/enumApi.ts#L599-L606

PR #366, opened 2023-01-16
fix generated enum value

trim21:
close https://github.com/oazapfts/oazapfts/issues/365

### DIFF
--- a/demo/petstore.json
+++ b/demo/petstore.json
@@ -1216,7 +1216,7 @@
           "size": {
             "description": "Size scale for pets",
             "enum": ["P", "M", "G", "0"],
-            "type": "number"
+            "type": "string"
           },
           "typeId": {
             "description": "integer test case for #349",
--- a/demo/enumApi.ts
+++ b/demo/enumApi.ts
@@ -591,10 +591,10 @@ export enum Status {
   $10Percent = "10percent",
 }
 export enum Size {
-  P = P,
-  M = M,
-  G = G,
-  $0 = 0,
+  P = "P",
+  M = "M",
+  G = "G",
+  $0 = "0",
 }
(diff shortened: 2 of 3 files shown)


demo/petstore.json
```diff
@@ -1216,7 +1216,7 @@
           "size": {
             "description": "Size scale for pets",
             "enum": ["P", "M", "G", "0"],
-            "type": "number"
```
fgnass, line 1219: I totally see your point in the other test-case. But here, I would expect it to use the index, when type is `number` and no explicit values are given. Otherwise it would also be a breaking change. Or am I missing something?

trim21: `"enum": ["P", "M", "G", "0"], type": "number"}` is not a valid json schema ...

trim21: And, in ts, enum type present it's value.

trim21: For a more complex example, you have a API take a enum as request body

```
export enum Size {
  P = 0,
  M = 1,
  G = 2,
  $0 = 3,
}

declare function fetch(s: Size): Promise<{}>
```

These pass type-checking:

```ts
fetch(1);
fetch(2);
fetch(3);
```

These don't

```ts
fetch('G');
fetch('P');
fetch('M');
```

fgnass:
The CI test run failed. I'm nowhere near a computer right now and will take a look later.

trim21:
@fgnass I forget to update test case, and ci of this PR didn't run because github think I'm new contributor of this repo, and you merge it without enable it...

I fixed it in https://github.com/oazapfts/oazapfts/pull/368 , you can mereg it to

fgnass:
Published as `oazapfts@4.5.2`

Issue #390, opened 2023-03-06
throw error with http headers


trim21:
Hi, I'm tring to catch error and get response header from `catch` error,

I can send a PR, are you open to this?

PR #391, opened 2023-03-06
Throw error with HTTP headers

trim21:
close https://github.com/oazapfts/oazapfts/issues/390

I have a use case like this:

```ts
try {
  await ok(...);
} catch (e) {
  if (e instanceof HttpError) {
    console.log(e.headers.get("x-request-id"));
  }

  throw e;
}
```

`oazapfts.fetchText` and `oazapfts.fetchJson` already return response with headers, so it doesn't change generated code, only runtime code and types.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,4 @@
-import { ApiResponse } from "./runtime";
+import { ApiResponse, WithHeaders } from "./runtime";
 
 /**
  * Type to access a response's data property for a given status.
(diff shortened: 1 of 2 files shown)


Xiphe:
Sorry for the late response. I'm kinda overloaded at the moment. Really like this change 👍 would you mind adding a unit-test for this? Not because the functionality is fragile but to document that this is intended and make sure it stays functional with future refactorings.

Xiphe:
pre-released as `oazapfts@4.6.0-throw-with-headers.1`

trim21:
@Xiphe tests added

Xiphe:
Cool!

trim21:
any new stable release?

Xiphe:
on it

Issue #419, opened 2023-05-08
v4.7.0 broke code generation


cvereterra:
Hi guys,

Mea culpa, [my PR](https://github.com/oazapfts/oazapfts/commit/57a4e01df039164dbdcb8c5ee25663c2b98f74da) broke code generation, the tests didn't catch this.
 
@Xiphe I suggest rolling back to ``v4.6.0``. In the meantime, I will try to fix this and add a PR with a test.

Given this endpoint definition 
```json
"paths": {
    "/healthcheck": {
      "get": {
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "application/json": {
                "schema": {
                  "properties": {
                    "message": {
                      "type": "string",
                      "example": "I'm health"
                    }
                  },
                  "required": ["message"]
                }
              }
            }
          }
        }
      }
    },
 ...
}
```

The response's schema should be:

```typescript
{
    message: string;
}
```

Before v4.7.0, oazapfts generated

```typescript
export function getHealthcheck(opts?: Oazapfts.RequestOpts) {
    return oazapfts.fetchJson<{
        status: 200;
        data: {
            message: string;
        };
    }>("/healthcheck", {
        ...opts
    });
}
```

Whereas now, it doesn't type the response
```typescript
export function getHealthcheck(opts?: Oazapfts.RequestOpts) {
    return oazapfts.fetchJson<{
        status: 200;
        data: {};
    }>("/healthcheck", {
        ...opts
    });
}
```

cvereterra:
Narrowed it down to the filtering done here in generate.ts (line 741). `onlyMode` is passed as `readOnly` since it's the result of a `GET`. The schema is not flagged as `readOnly` by `isSchemaReadOnly`, so it get's filtered out.


```typescript
 /**
   * Recursively creates a type literal with the given props.
   */
  getTypeFromProperties(
    props: {
      [prop: string]: SchemaObject | OpenAPIV3.ReferenceObject;
    },
    required?: string[],
    additionalProperties?:
      | boolean
      | OpenAPIV3.SchemaObject
      | OpenAPIV3.ReferenceObject,
    onlyMode?: OnlyMode
  ): ts.TypeLiteralNode {
    const members: ts.TypeElement[] = Object.keys(props)
      .filter((name) => {
        const schema = props[name];
        if (!onlyMode)
          return (
            !this.isSchemaReadOnly(schema) && !this.isSchemaWriteOnly(schema)
          );
[...]

PR #420, opened 2023-05-08
Fix 419

cvereterra:
Addresses https://github.com/oazapfts/oazapfts/issues/419

Added a check to see if we have filtered out all props

```typescript
// Check if any of the props are readOnly or writeOnly schemas
const propertyNames = Object.keys(props);
const filteredPropertyNames = propertyNames.filter((name) => {
  const schema = props[name];
  if (!onlyMode)
    return (
      !this.isSchemaReadOnly(schema) && !this.isSchemaWriteOnly(schema)
    );
  if (onlyMode === "readOnly") return this.isSchemaReadOnly(schema);
  else if (onlyMode === "writeOnly") return this.isSchemaWriteOnly(schema);
});
// By filtering by readOnly/writeOnly props, we may have filtered out all props in schemas
const hasFilteredAllProps = filteredPropertyNames.length === 0;
const names = hasFilteredAllProps ? propertyNames : filteredPropertyNames;

```


Xiphe:
Awesome! Thanks for notifying and fixing this so fast!

Issue #446, opened 2023-07-28
Read/Write-Only bugs in generated code when referenced in collections


ehvattum:
Hello,

It seems the readonly and writeonly support introduced in 57a4e01df03 have broken some cases of code generation.

We have a type `Product` with a readonly `id` property.
We have a type `PagedListOfProduct` with a property `items` that is a collection of `Product`s.

When using oazapfts version 4.7.3, the generated PagedListOfProduct type references the _default_ `Product` type, while I would expect it to reference the generated `ProductRead` type instead.

Schema:

```json
{
    "openapi": "3.0.0",    
    "servers": [
        {
            "url": "localhost:5001"
        }
    ],
    "paths": {
        "/api/products": {
            "get": {
                "tags": [
                    "Products"
                ],
                "operationId": "Products_GetAll",
                "parameters": [
                    {
                        "name": "PageIndex",
                        "in": "query",
                        "schema": {
                            "type": "integer",
                            "format": "int32",
                            "default": 0,
                            "maximum": 2147483647,
                            "minimum": 0
                        }
                    },
                    {
                        "name": "PageSize",
                        "in": "query",
                        "schema": {
                            "type": "integer",
                            "format": "int32",
                            "default": 15,
                            "maximum": 2147483647,
                            "minimum": 1
                        }
                    }
                ],
                "responses": {
                    "200": {
                        "description": "",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/PagedListOfProduct"
                                }
                            }
                        }
                    }
                }
            }
        }
    },
    "components": {
        "schemas": {
            "PagedListOfProduct": {
                "type": "object",
                "additionalProperties": false,
                "required": [
                    "items",
                    "totalPages",
                    "pageIndex",
                    "pageSize",
                    "totalCount"
                ],
                "properties": {
                    "items": {
                        "type": "array",
                        "items": {
                            "$ref": "#/components/schemas/Product"
                        }
                    },
                    "totalPages": {
                        "type": "integer",
                        "format": "int32"
                    },
                    "pageIndex": {
                        "type": "integer",
                        "format": "int32"
                    },
                    "pageSize": {
                        "type": "integer",
                        "format": "int32"
                    },
                    "totalCount": {
                        "type": "integer",
                        "format": "int32"
                    }
                }
            },
            "Product": {
                "type": "object",
                "additionalProperties": false,
                "required": [
                    "id",
                    "name",
                    "description",
[...]

PR #450, opened 2023-07-31
add --mergeReadWriteOnly flag

Xiphe:
in order to opt-out of the read-/writeOnly split behaviour

ref https://github.com/oazapfts/oazapfts/issues/446

### DIFF
--- a/src/codegen/cli.ts
+++ b/src/codegen/cli.ts
@@ -10,7 +10,7 @@ const argv = minimist(process.argv.slice(2), {
     i: "include",
     e: "exclude",
   },
-  boolean: ["optimistic", "useEnumType"],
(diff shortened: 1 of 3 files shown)


Xiphe:
Pre-released as `oazapfts@4.7.4-merge-read-write.1` 

@ehvattum this is a quickfix/workaround that should allow you to opt back into the old behavior. Can you give this a try?

The root fix for https://github.com/oazapfts/oazapfts/issues/446 is probably a little more complicated and might lead to some discussions.

ehvattum:
This is a great way to restore the previous behavior.

Xiphe:
Cool. Gonna merge this now as I think it's a good addition even when we fix the root cause.

For context: It seems to me as if the readonly/writeonly pattern is also not properly implemented by some other tools, so having an option to opt out of this might be useful.

Issue #382, opened 2023-02-14
Support `readOnly` and `writeOnly`


olivierbeaulieu:
oazapfts doesn't seem to support these two properties from OAS. It would be a hugely helpful feature to support, as today we need to create distinct schemas for requests & responses.

Example: Given the OpenAPI spec below, the generated `postThing` function should have `read_only_prop` should only be present in the response, and `write_only_prop` should only be present in the request.

```json
{
  "info": {
    "title": "Example"
  },
  "openapi": "3.0.2",
  "servers": [
    {
      "url": "https://example.com"
    }
  ],
  "paths": {
    "/post-thing": {
      "post": {
        "operationId": "post_thing",
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/ExampleSchema"
              }
            }
          },
          "required": true
        },
        "responses": {
          "201": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/ExampleSchema"
                }
              }
            },
            "description": "Success"
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "ExampleSchema": {
        "properties": {
          "read_only_prop": {
            "type": "string",
            "readOnly": true
          },
          "write_only_prop": {
            "type": "boolean",
            "writeOnly": true
          }
        },
        "type": "object"
      }
    }
  }
}
```

Xiphe:
That's an interesting one! Agree that it would be cool if we could support this somehow.

In TypeScript these would still need to be two different entities since the language does not know about the context in which the types are used. This means we need to come up with a naming schema for those and in the generated code you'd still have something like this:

```ts
type ExampleSchemaReq {
  write_only_prop: boolean
}
type ExampleSchemaRes {
   read_only_prop: boolean
}
```

Any suggestions for naming? And anyone keen to implement this?

jonkoops:
Interesting, perhaps we could solve this with generics somehow? Or perhaps that would only make things more complicated.

As for naming, probably best to stick to the schema naming convention as much as possible. So perhaps `ExampleSchemaRead` and `ExampleSchemaWrite`?

jonkoops:
Alternatively, we could keep the type representation for the schema in a more complex type (closer to spec) that contains all the information for the fields (such as `type`, `writeOnly`, `readOnly`) and provide utility types that can 'serialize' this to a 'plain' type when returned from methods.

```ts
[...]

PR #451, opened 2023-07-31
Handle nested objects and $refs in read-/writeonly types

Xiphe:
Supersedes: https://github.com/oazapfts/oazapfts/pull/450

The existing implementation for [handling read-/writeonly](https://github.com/oazapfts/oazapfts/issues/382) ignored `array`, `$ref` and other nested types, which caused some properties to become missing on those nested objects.

This fix now always creates all applicable read/write extensions for a type and takes nested/linked types into account when determining if a type has readonly or writeonly qualities.

Also keeps the `--mergeReadWriteOnly` flag from #450 around to support opting out of the the behavior altogether.

@cvereterra would you mind having a look on this?

Also PTAL/FYI @fgnass @jonkoops @ehvattum

### DIFF
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -753,11 +753,8 @@ export default class ApiGenerator {
           return !isReadOnly && !isWriteOnly;
       }
     });
-    // By filtering by readOnly/writeOnly props, we may have filtered out all props in schemas
-    const hasFilteredAllProps = filteredPropertyNames.length === 0;
-    const names = hasFilteredAllProps ? propertyNames : filteredPropertyNames;
 
-    const members: ts.TypeElement[] = names.map((name) => {
+    const members: ts.TypeElement[] = filteredPropertyNames.map((name) => {
       const schema = props[name];
       const isRequired = required && required.includes(name);
       let type = this.getTypeFromSchema(schema, name, onlyMode);
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -289,16 +289,15 @@ export default class ApiGenerator {
 
   // Collect the types of all referenced schemas so we can export them later
   // Referenced schemas can be pointing at the following versions:
-  // - "none": The regular type/interface e.g. ExampleSchema
+  // - "base": The regular type/interface e.g. ExampleSchema
   // - "readOnly": The readOnly version e.g. ExampleSchemaRead
   // - "writeOnly": The writeOnly version e.g. ExampleSchemaWrite
   refs: Record<
     string,
     {
-      [k in OnlyMode | "none"]:
-        | ts.TypeReferenceNode
-        | ts.InterfaceDeclaration
-        | undefined;
+      base: ts.TypeReferenceNode;
+      readOnly?: ts.TypeReferenceNode;
+      writeOnly?: ts.TypeReferenceNode;
     }
   > = {};
 
@@ -364,15 +363,6 @@ export default class ApiGenerator {
   getRefAlias(obj: OpenAPIV3.ReferenceObject, onlyMode?: OnlyMode) {
     const { $ref } = obj;
     if (!this.refs[$ref]) {
-      this.refs[$ref] = {
-        none: undefined,
-        readOnly: undefined,
-        writeOnly: undefined,
-      };
-    }
-    let ref = this.refs[$ref][onlyMode ?? "none"];
-
-    if (!this.refs[$ref]["none"]) {
       const schema = this.resolve<SchemaObject>(obj);
       const name = schema.title || getRefName($ref);
       const identifier = toIdentifier(name, true);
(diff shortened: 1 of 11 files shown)


ehvattum:
I have tested a little, and it seems there are other bugs introduced now.
When generating without using the new flag; 
If a type `A` is expressed as `allOf B`, this new PR causes the generated type A to not extend B.
This happens if any nested property on type `A` is readOnly. The interface for `A` does not extend the type of `B`.

Xiphe:
Thanks for testing @ehvattum 😅 Seems as if I'll need to iterate on this some more (tomorrow at earliest)

Hypothetically, the `allOf` example should create something Like this, shouldn't it?

```ts
interface B {
  normalPropB: string;
}
interface BRead extends B {
  readonlyPropB: number;
}
interface A {
  normalPropA: string;
}
interface ARead extends A, BRead {}
```

ehvattum:
Given the spec
```json
{
    "x-generator": "NSwag v13.16.1.0 (NJsonSchema v10.7.2.0 (Newtonsoft.Json v13.0.0.0))",
    "openapi": "3.0.0",
    "info": {
        "title": "API",
        "version": "13.0.0"
    },
    "paths": {
        "/api/users/me": {
            "get": {
                "tags": [
                    "Users"
                ],
                "operationId": "Users_GetMe",
                "responses": {
                    "200": {
                        "description": "",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/User"
                                }
                            }
                        }
                    }
                }
            }
        }
    },
    "components": {
        "schemas": {
            "CompactUser": {
                "type": "object",
                "additionalProperties": false,
                "required": [
                    "id",
                    "email",
                    "displayName"
                ],
                "properties": {
                    "id": {
                        "type": "string",
                        "format": "guid",
                        "minLength": 1
                    },
                    "email": {
                        "type": "string",
                        "format": "email",
                        "minLength": 1
                    },
                    "displayName": {
                        "type": "string",
                        "minLength": 1
                    }
                }
            },
            "PendingMembership": {
                "type": "object",
                "additionalProperties": false,
                "required": [
                    "id"
                ],
                "properties": {
                    "id": {
                        "type": "string",
                        "readOnly": true,
                        "format": "guid",
                        "minLength": 1
                    }
                }
            },
            "User": {
                "allOf": [
                    {
                        "$ref": "#/components/schemas/CompactUser"
                    },
                    {
                        "type": "object",
                        "additionalProperties": false,
                        "required": [
                            "externalID",
                            "pendingMemberships"
                        ],
                        "properties": {
                            "externalID": {
                                "type": "string",
                                "minLength": 1
                            },
                            "pendingMemberships": {
                                "type": "array",
                                "items": {
                                    "$ref": "#/components/schemas/PendingMembership"
                                }
                            }
                        }
                    }
                ]
            }
        }
    }
}

```
We get the following code generated.

```ts
export type CompactUser = {
    id: string;
    email: string;
    displayName: string;
};
export interface PendingMembership {
    id: string;
}
export interface PendingMembershipRead extends PendingMembership {
    id: string;
}
export interface User {
}
export interface UserRead extends User {
}
export function usersGetMe(opts?: Oazapfts.RequestOpts) {
    return oazapfts.fetchJson<{
        status: 200;
        data: UserRead;
    }>("/api/users/me", {
        ...opts
    });
}
```

The issue is that User is not extending CompactUser.

Xiphe:
So It turned out that the whole `interface Base {}; interface BaseWrite extends Base {}`; approach caused a big amount of complexity when I tried to make it work with nested components, `allOf`, `oneOf` etc.

I therefore decided to simplify the approach to not try to extend anything and just duplicate the props on the base/read/write variants.

Would be interested in your feedback to this @jonkoops as the current implementation is based on [your suggestion](https://github.com/oazapfts/oazapfts/issues/382#issuecomment-1431181183)

src/codegen/generate.ts
```diff
@@ -364,15 +363,6 @@ export default class ApiGenerator {
   getRefAlias(obj: OpenAPIV3.ReferenceObject, onlyMode?: OnlyMode) {
```
Xiphe, line 363: The changes in this method essentially make it so that we always create all possible permutations of a referenced type (base/read/write).
Prior to this we were a little more regressive and only created the read/write types when the ref actually was used in a context where they would be needed.

src/codegen/generate.ts
```diff
@@ -706,36 +681,46 @@ export default class ApiGenerator {
     return type;
   }
 
-  /**
-   * Check if schema is readOnly or has readOnly properties
-   */
-  isSchemaReadOnly(schema: SchemaObject | OpenAPIV3.ReferenceObject) {
+  checkSchemaOnlyMode(
```
Xiphe, line 684: merged `isSchemaReadOnly` and `isSchemaWriteOnly` and will now consider a schema to have write/read only qualities when any deeply nested prop has the feature

src/codegen/generate.ts
```diff
@@ -756,12 +741,17 @@ export default class ApiGenerator {
...
+      const isWriteOnly = this.checkSchemaOnlyMode(schema, "writeOnly", false);
+
+      switch (onlyMode) {
+        case "readOnly":
+          return isReadOnly || !isWriteOnly;
+        case "writeOnly":
+          return isWriteOnly || !isReadOnly;
+        default:
+          return !isReadOnly && !isWriteOnly;
+      }
```
Xiphe, line 754: now essentially only filtering out read+write props for base types, write props for read types and read props for write types

src/codegen/generate.ts
```diff
@@ -385,85 +375,67 @@ export default class ApiGenerator {
...
 
     // If not ref fallback to the regular reference
-    return ref ?? this.refs[$ref]["none"];
+    return this.refs[$ref][onlyMode || "base"] ?? this.refs[$ref].base;
   }
 
   getUnionType(
     variants: (OpenAPIV3.ReferenceObject | SchemaObject)[],
     discriminator?: OpenAPIV3.DiscriminatorObject,
+    onlyMode?: OnlyMode,
```
Xiphe, line 438: the `onlyMode` is now passed around much more and more deeply so that when we create a Parent type in "writeOnly" mode, and that one has a $ref for example, we'll try attaching the writeOnly version of that ref.

Xiphe:
Pre-released this as `oazapfts@4.7.4-read-write-fixes.1`

Xiphe:
While at it @bdm-k made us aware of another related issue https://github.com/oazapfts/oazapfts/issues/453 I'm gonna try to address this in this PR.

Xiphe:
New version addressing #453: `oazapfts@4.8.0-read-write-fixes.2`

Xiphe:
I'm gonna be on vacation until August 28th. Gonna merge this when I'm back.

Issue #452, opened 2023-08-04
`Content-Type` Header Is not updated


kmirojo:
Good day 👋 
I'm using `oazapfts@4.7.4` to implement my fetch calls from an `Open API` specification file in a typescript client.

I need to configure the `oazapfts` request to send whole body unicode/utf-8, but when I try to do it, it is overrriten by the default headers configuration.

In my API configuration file I'm setting the header in multiple ways:
```ts
defaults.headers = {
      "Content-Type": "application/json; charset=UTF-8",
};
//
defaults.headers["Content-Type"] = application/json; charset=UTF-8";
```

But every time the fetch call header looks like `Content-Type: "application/json"`.

Am I doing something wrong? (most likely)
Or is there no way to set the header `Content-Type` header according to my needs?

Xiphe:
Hi @kmirojo 

You're right, it seems as if the Content-Type header currently can not be overwritten: https://github.com/oazapfts/oazapfts/blob/main/src/runtime/index.ts#L114-L115

I'm not entirely sure how to best solve this. But I tend to say we should switch those lines. Any feedback/input very welcome

kmirojo:
Thanks a lot for taking the time to look at this @Xiphe 🙏 .
We will be attentive to the implementation of the suggested solution whenever it's possible or if you require more information from our side,

Xiphe:
Happy to review a PR on this 👍

kmirojo:
Thanks @Xiphe 
I'll create a PR for this issue, but we were wondering, how soon would you guys publish another release if the PR meets the acceptance criteria.

Xiphe:
We're utilizing [semantic release](https://semantic-release.gitbook.io/semantic-release/). So when the change is in `main` there will be a release.

And given the small non-intrusive nature of the change I see no blockers that would delay a swift merge of the PR.

PR #456, opened 2023-08-10
Customize 'Content-Type' header

kmirojo:
This Pull Request attempts to allow the `Content-Type` header to be customized.
Here's a little more context about the reason for this proposal: https://github.com/oazapfts/oazapfts/issues/452

Hi @Xiphe 
In our conversation to allow this (https://github.com/oazapfts/oazapfts/issues/452), you commented:
> Gave this another thought. I think just switching out the lines should be treated as a breaking change ...

But I wasn't sure if you were just not convinced about switching these lines. 
I gave some thoughts about how these changes could affect negatively another user but I think it would only be an issue if they have an unnoticed modification of the `Content-Type` header. But of course, I could totally be missing some scenarios.

Please let me know if you think we should go for a more complex approach on this or if you have any other thoughts about this.

### DIFF
--- a/src/runtime/index.ts
+++ b/src/runtime/index.ts
@@ -100,15 +100,15 @@ export function runtime(defaults: RequestOpts) {
     return res;
   }
 
-  function hasJsonContentType(contentTypeHeader: string) {
-    console.log("hasJsonContentType");
+  /* TODO: maybe we want to get rid of this checks (ensureJsonContentType & ensureFormContentType)
+  in a future major release (ref https://github.com/oazapfts/oazapfts/pull/456) */
+  function ensureJsonContentType(contentTypeHeader: string) {
     return contentTypeHeader?.match(/\bjson\b/i)
       ? contentTypeHeader
       : "application/json";
   }
 
-  function hasFormContentType(contentTypeHeader: string) {
-    console.log("hasFormContentType");
+  function ensureFormContentType(contentTypeHeader: string) {
     return contentTypeHeader?.startsWith("application/x-www-form-urlencoded")
       ? contentTypeHeader
       : "application/x-www-form-urlencoded";
@@ -126,7 +126,9 @@ export function runtime(defaults: RequestOpts) {
         ...(body != null && { body: JSON.stringify(body) }),
         headers: {
           ...headers,
-          "Content-Type": hasJsonContentType(String(headers?.["Content-Type"])),
+          "Content-Type": ensureJsonContentType(
+            String(headers?.["Content-Type"]),
+          ),
         },
       };
     },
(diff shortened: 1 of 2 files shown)


Xiphe:
You're probably right with your assessment. But you never know what kind of abstractions people build some might re-use a header object originally constructed for formData requests or s.th. like that. 
I agree that the chances for this are really low but I don't want to accidentally break stuff with such a minor change. 
The alternative would be to release the change as v5 but that also seems a little out of scale.

So please let's check if the overwriting header looks like a json header and only then allow the overwrite.

```ts
headers: {
  ...headers,
  "Content-Type":  headers?.["Content-Type"]?.match(/\bjson\b/i) ? headers["Content-Type"] : "application/json",
},
```

Also while at it, let's do the same treatment to the [`form`](https://github.com/oazapfts/oazapfts/blob/4cf87299f24f96b572e28982e381e8a55953da59/src/runtime/index.ts#L120-L129) method by checking for `headers?.["Content-Type"]?.startsWith('application/x-www-form-urlencoded')`.

When feasible let's also add a small unit test for this and add a in-code comment to remove the checks on the next major release.

Let me know when you want to me take over some of the parts.

kmirojo:
Hey @Xiphe, I've just pushed some changes and included one test for the work I'm doing, please LMK your thoughts. 

> When feasible let's also add a small unit test for this and add a **in-code comment to remove the checks** on the next major release.

I did not quite get the `in-code comment` part, are you referring to a regular comment (`//`) explaining the changes or do I need some kind of prefix? 😬

Xiphe:
Awesome! Yeah, I meant something like `/* TODO: maybe we want to get rid of this check in a future major release (ref https://github.com/oazapfts/oazapfts/pull/456) */` comment

src/runtime/index.ts
```diff
@@ -100,6 +100,20 @@ export function runtime(defaults: RequestOpts) {
     return res;
   }
 
+  function hasJsonContentType(contentTypeHeader: string) {
+    console.log("hasJsonContentType");
```
Xiphe, line 104: Can you please remove the logger

src/runtime/index.ts
```diff
@@ -100,6 +100,20 @@ export function runtime(defaults: RequestOpts) {
...
+    return contentTypeHeader?.match(/\bjson\b/i)
+      ? contentTypeHeader
+      : "application/json";
+  }
+
+  function hasFormContentType(contentTypeHeader: string) {
+    console.log("hasFormContentType");
```
Xiphe, line 111: and this one

src/runtime/index.ts
```diff
@@ -100,6 +100,20 @@ export function runtime(defaults: RequestOpts) {
     return res;
   }
 
+  function hasJsonContentType(contentTypeHeader: string) {
```
Xiphe, line 103: This is nit-picky but i'd expect a `has...` function to return a `boolean` can we name it `ensureJsonContentType` instead or similar?

Xiphe:
Thank you so much for putting in the afford! Already looks good to me, just left some tiny comments

Xiphe:
Thank you for the contribution 👍

PR #462, opened 2023-08-16
Add prefixItems to type generation

### LINKED ISSUE #461: Schema with prefixItems translates to any type
I original reported this at the rtk-toolkit repo reduxjs/redux-toolkit#3648 but now dove in a little and found that the issue is with oazapfts. In short, I have a property in an OpenAPI schema that is defined as an array of two `anyOf`s of type `integer` or `number` with minimum and maximum length of 2. I would expect this to lead to type of `[number, number]` but instead I receive `any`. I glanced over the issues #422 and #417 since they are also related to `allOf` but it wasnt obvious that they are the same issue.

## Reproducible example
Given this openapi schema.
```json 
{
  "openapi": "3.1.0",
  "info": {
    "title": "Test",
    "description": "",
    "version": "1.0.0"
  },
  "paths": {
    "/test/": {
      "get": {
        "summary": "",
        "description": "",
        "operationId": "get_test",
        "parameters": [],
        "responses": {
          "200": {
            "description": "Successful Response",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/Coordinates"
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Coordinates": {
        "prefixItems": [
          {
            "anyOf": [
              {
                "type": "number"
              },
              {
                "type": "integer"
              }
            ],
            "title": "Lon"
          },
          {
            "anyOf": [
              {
                "type": "number"
              },
              {
                "type": "integer"
              }
            ],
            "title": "Lat"
          }
        ],
        "type": "array",
        "maxItems": 2,
        "minItems": 2
      }
    }
  }
}
```

I receive these type definitions from oazapfts

```ts
/**
 * Test
 * 1.0.0
 * DO NOT MODIFY - This file has been generated using oazapfts.
 * See https://www.npmjs.com/package/oazapfts
 */
import * as Oazapfts from "oazapfts/lib/runtime";
import * as QS from "oazapfts/lib/runtime/query";
export const defaults: Oazapfts.RequestOpts = {
    baseUrl: "/",
};
const oazapfts = Oazapfts.runtime(defaults);
export const servers = {};
export type Coordinates = any;
export function getTest(opts?: Oazapfts.RequestOpts) {
    return oazapfts.fetchJson<{
        status: 200;
        data: Coordinates;
    }>("/test/", {
        ...opts
    });
}
```

In L14 `type Coordinates = any` I would expect `type Coordinates = [number, number]`.

saschahofmann:
Closes #461

I simply added another if statement and added the prefixItems to the SchemaObject type.

Also would you like me to add a test?

Are you using prettier or something, mine seems to be on different settings.

### DIFF
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -51,6 +51,7 @@ type SchemaObject = OpenAPIV3.SchemaObject & {
   const?: unknown;
   "x-enumNames"?: string[];
   "x-enum-varnames"?: string[];
+  prefixItems?: (OpenAPIV3.ReferenceObject | SchemaObject)[];
 };
 
 /**
@@ -588,18 +589,10 @@ export default class ApiGenerator {
       // items -> array
       return factory.createArrayTypeNode(this.getTypeFromSchema(schema.items));
     }
-    if ("prefixItems" in schema) {
+    if ("prefixItems" in schema && schema.prefixItems) {
       // prefixItems -> typed tuple
-      // Because we are mixing OpenAPI 3.0 and OpenAPI 3.1 types prefixItems can be defined without items
-      // For some reason typescript assumes schema to be NonArraySchemaObject after the above if statement so we cast it back
-      // Also the definition of prefixItems in openapi-types seems to be wrong so we cast again
-      const prefixItems = (schema as unknown as OpenAPIV3.ArraySchemaObject)
-        .prefixItems as unknown as (
-        | OpenAPIV3.SchemaObject
-        | OpenAPIV3.ReferenceObject
-      )[];
       return factory.createTupleTypeNode(
-        prefixItems.map((schema) => this.getTypeFromSchema(schema)),
+        schema.prefixItems.map((schema) => this.getTypeFromSchema(schema)),
       );
     }
     if (schema.properties || schema.additionalProperties) {
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -588,6 +588,20 @@ export default class ApiGenerator {
       // items -> array
       return factory.createArrayTypeNode(this.getTypeFromSchema(schema.items));
     }
+    if ("prefixItems" in schema) {
+      // prefixItems -> typed tuple
+      // Because we are mixing OpenAPI 3.0 and OpenAPI 3.1 types prefixItems can be defined without items
+      // For some reason typescript assumes schema to be NonArraySchemaObject after the above if statement so we cast it back
+      // Also the definition of prefixItems in openapi-types seems to be wrong so we cast again
+      const prefixItems = (schema as unknown as OpenAPIV3.ArraySchemaObject)
+        .prefixItems as unknown as (
+        | OpenAPIV3.SchemaObject
+        | OpenAPIV3.ReferenceObject
+      )[];
+      return factory.createTupleTypeNode(
+        prefixItems.map((schema) => this.getTypeFromSchema(schema)),
+      );
+    }
     if (schema.properties || schema.additionalProperties) {
       // properties -> literal type
       return this.getTypeFromProperties(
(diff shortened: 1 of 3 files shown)


Xiphe:
Hey @saschahofmann thanks for reporting the issue and also providing the PR

Would you mind adding a test-case for this?

saschahofmann:
Sure thing. I was looking at *generate.test.ts* and saw that the produces types arent checked for yet. Do you have an example test somewhere else for this?

Xiphe:
I think this could be tested in `src/codegen/index.test.ts` by asserting the presence of the correct type in the string output of `generate`
Or you could add a reproduction to `demo/petstore.json` and do something like this:

```ts
// demo/api.test.ts

test('Coordinates is not any', () => {
  // @ts-expect-error
  const coords: api.Coordinates = 'not-coordinates';
});
```

I think i'd prefer the first option but both should be fine.

src/codegen/generate.ts
```diff
@@ -588,6 +588,20 @@ export default class ApiGenerator {
...
+    if ("prefixItems" in schema) {
+      // prefixItems -> typed tuple
+      // Because we are mixing OpenAPI 3.0 and OpenAPI 3.1 types prefixItems can be defined without items
+      // For some reason typescript assumes schema to be NonArraySchemaObject after the above if statement so we cast it back
+      // Also the definition of prefixItems in openapi-types seems to be wrong so we cast again
+      const prefixItems = (schema as unknown as OpenAPIV3.ArraySchemaObject)
+        .prefixItems as unknown as (
+        | OpenAPIV3.SchemaObject
+        | OpenAPIV3.ReferenceObject
+      )[];
```
Xiphe, line 600: On my end this works fine:

```ts
const prefixItems = schema.prefixItems as (
  | OpenAPIV3.SchemaObject
  | OpenAPIV3.ReferenceObject
)[];
```

Ideally we'd have typeguards here but I haven't found existing implementations for a `isSchemaObject` and `isReferenceObject` and I'd not bother implementing those ourselves just for this case.

Xiphe: great solution 👍

saschahofmann:
Alright I missed that *index.test.ts*, I added an extra *prefixItems.json* and just check that the type is created correctly.

Xiphe:
LGTM 🎉 should we merge this?

saschahofmann:
Lets do it!

src/codegen/generate.ts
```diff
@@ -588,6 +589,12 @@ export default class ApiGenerator {
       // items -> array
       return factory.createArrayTypeNode(this.getTypeFromSchema(schema.items));
     }
+    if ("prefixItems" in schema && schema.prefixItems) {
+      // prefixItems -> typed tuple
+      return factory.createTupleTypeNode(
+        schema.prefixItems.map((schema) => this.getTypeFromSchema(schema)),
+      );
+    }
```
felixfbecker, line 597: Shouldn't this check `maxItems` and `items` so that if `prefixItems` is truly used to only describe a prefix, it becomes an open-ended tuple like `[...TPrefixItems, ...TItems]`?

saschahofmann: To be honest, I didnt really understand what `prefixItems` are supposed to  do (just checked [here](https://json-schema.org/understanding-json-schema/reference/array?highlight=prefixitems#additional-items) but I think you're right, the result should add potential additional items. Currently, `items` and `prefixItems` cannot be used together. I doubt it's too hard to do though?

Xiphe: Should we treat this as a bug? Is there an issue for this?

Xiphe: For context, I'm just getting back from a 6weeks time-off and don't have the time to understand in detail what this is about. If this is something that will affect others, can one of you please open an issue?

PR #470, opened 2023-08-28
fix: handler functions' argument type

### LINKED ISSUE #460: `handle` helper returns `never` type for non-array objects
#### Description

I have an OpenAPI spec that contains endpoints similar to 

```yaml
paths:
  /cats:
    get:
      operationId: listCats
      responses:
        "200":
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: "#/components/schemas/Cat"

  /cats/{id}:
    get:
      operationId: getCat
      responses:
        "200":
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Cat"
```

The type signatures of the generated functions corresponding to these operations are

```ts
function listCats(opts?: RequestOpts | undefined): Promise<WithHeaders<{
    status: 200;
    data: Cat[];
}>>

function getCat(id: number, opts?: RequestOpts | undefined): Promise<WithHeaders<{
    status: 200;
    data: Cat;
}>>
```

When using the `handle` helper like so

```ts
const cats = await handle(listCats(), {
  200: (cats) => cats,
}

const cat = await handle(getCat(), {
  200: (cat) => cat
}
```
we expect the types of `cats` and `cat` to be `Cat[]` and `Cat` respectively.

The issue I am having is that while the type of `cats` is inferred correctly, the type of `cat` becomes `never`. This is only the case for non-array types.

#### Comments

Looking into the type declarations used by `handle`, we find this generic type that is used to determine the return type of `handle`.

```ts
export type FunctionReturnType<T> = T extends (args: any[]) => any ? ReturnType<T> : never;
```

I'm far from a typescript expert, but as I understand it, this will equate to "if the handler type extends `(args: any[]) => any`, then use the return type of this handler and otherwise use `never`". The key detail here is the type of `args`, `any[]`. This will only match arrays, and hence, we get a `never` type for handlers that return values that are not arrays (such as singular objects). Changing this type to

```ts
export type FunctionReturnType<T> = T extends (args: any) => any ? ReturnType<T> : never;
```

appears to resolve the issue. The type of `cat` is now correctly `Cat`, and the type of handlers that have no return values remains `never`, as it should.

I can submit a PR with this fix, but it would be appreciated if someone can lay eyes on this to make sure I'm not missing something.

eric-carlsson:
#### Description
* Fixes the type used to match handler functions inside `handle`, [FunctionReturnType](https://github.com/oazapfts/oazapfts/blob/4fafe0028f0484153c9f458aba5e0fb5e0f4f4ae/src/index.ts#L21-L23), so that it correctly matches handlers with any arguments.
* Add a test to assert the return type of `handle`. This test is non-complete, since technically any type that `api.Pet` can partially satisfy will pass, but it's the best I can come up with since `never` is a subtype to all types.

#### Related issues
- Closes #460

### DIFF
--- a/demo/api.test.ts
+++ b/demo/api.test.ts
@@ -277,6 +277,17 @@ describe("handle", () => {
     const promise = handle(api.updatePet({ name: "Gizmo", photoUrls: [] }), {});
     await expect(promise).rejects.toHaveProperty("status", 204);
   });
+
+  it("should type response as Pet", async () => {
+    const res = await handle(api.getPetById(1), {
+      200(res) {
+        return res;
+      },
+    });
+
+    ({}) as api.Pet satisfies typeof res;
+    expect(res).toBeDefined();
+  });
 });
 
 describe("okify", () => {
--- a/src/index.ts
+++ b/src/index.ts
@@ -18,7 +18,7 @@ export type ResponseHandler<T extends ApiResponse> = {
   default?: (status: number, data: any) => any;
 };
 
-export type FunctionReturnType<T> = T extends (args: any[]) => any
+export type FunctionReturnType<T> = T extends (...args: any[]) => any
   ? ReturnType<T>
   : never;
 


eric-carlsson:
I updated the test from [the issue](https://github.com/oazapfts/oazapfts/issues/460#issuecomment-1695562393) to catch a broader group of potential type issues, but it's not fully complete. There are cases where the test will pass even if the returned type does not fully satisfy the generated model type. If anyone has any ideas on how this can be done better I'm all ears.

Xiphe:
Awesome! Thank you for the contribution!

PR #473, opened 2023-08-31
Support discriminator with allOf

### LINKED ISSUE #422: allOf with inheritance
I'd like to combine inheritance and the discriminator in my API, as this describes: https://redocly.com/docs/resources/discriminator/#allof-for-inheritance. It looks like oazapfts only supports using the discriminator with oneOf though. Is that correct? Is there an alternative?

Here's a rephrasing of the example from the link:
```
Vehicle:
  type: object
  description: Vehicle
  discriminator:
    propertyName: powerSource
    mapping:
      electricity: '#/components/schemas/ElectricVehicle'
      gasoline: '#/components/schemas/FueledVehicle'
      human-energy: '#/components/schemas/PedaledVehicle'
  properties:
    vehicleType:
      description: The type of vehicle.
      type: string
      example: bicycle
    idealTerrain:
      type: string
      example: roads
    powerSource:
      description: How is the vehicle powered.
      type: string
      example: pedaling
PedaledVehicle:
  allOf:
    - $ref: '#/components/schemas/Vehicle'
    - type: object
      description: Pedaled Vehicle
      properties:
        topSpeed:
          description: The top speed in kilometers per hour rounded to the nearest integer.
          type: integer
          example: 83
        range:
          description: The 95th percentile range of a trip in kilometers.
          type: integer
          example: 100

### PR DESCRIPTION
**Pull request by bdm-k**
Closes #422 

This PR adds support for the discriminator used in conjunction with `allOf` constructor. See the added tests for an example. Several key concepts are introduced in this PR:

1. `metadata` property: A new property called `metadata` has been added to the `SchemaObject`. This property contains information added by oazapft. As of now, only the `selfRef` is present in the `SchemaMetadata` type, but other properties may be added in the future.
2. parent schema: The term "parent schema" appears in several places. In this context, it refers to a schema that has a `discriminator` property which is neither used in conjunction with `oneOf` nor `anyOf`. This is also described in the comment of `preprocessSchemas`.
3. `ignoreDiscriminator` parameter: In order to support the discriminator used with `allOf`, it's important to determine whether a reference to a parent schema is pointing to the parent schema itself or the union of its alternative (child) schemas. The type alias for a parent schema itself can be obtained by passing `true` to the `ignoreDiscriminator`, which is a newly added parameter in `getRefAlias`.

### DIFF
--- a/src/codegen/__fixtures__/allOf.json
+++ b/src/codegen/__fixtures__/allOf.json
@@ -40,6 +40,37 @@
           }
         }
       }
+    },
+    "/pet/{pet_id}": {
+      "get": {
+        "tags": ["pet"],
+        "summary": "Find pet by ID",
+        "description": "Returns a single pet",
+        "operationId": "getPetById",
+        "parameters": [
+          {
+            "name": "pet_id",
+            "in": "path",
+            "description": "ID of pet to return",
+            "required": true,
+            "schema": {
+              "type": "integer"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "successful operation",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Pet"
+                }
+              }
+            }
+          }
+        }
+      }
     }
   },
   "components": {
@@ -74,6 +105,66 @@
             "type": "integer"
           }
         }
+      },
+      "Pet": {
+        "type": "object",
+        "properties": {
+          "petType": {
+            "type": "string"
+          }
+        },
+        "required": ["petType"],
+        "discriminator": {
+          "propertyName": "petType",
+          "mapping": {
+            "dog": "#/components/schemas/Dog"
+          }
+        }
+      },
+      "Cat": {
+        "allOf": [
+          {
+            "$ref": "#/components/schemas/Pet"
+          },
+          {
+            "type": "object",
+            "properties": {
+              "name": {
+                "type": "string"
+              }
+            }
+          }
+        ]
+      },
+      "Dog": {
+        "allOf": [
+          {
+            "$ref": "#/components/schemas/Pet"
+          },
+          {
+            "type": "object",
+            "properties": {
+              "bark": {
+                "type": "string"
+              }
+            }
+          }
+        ]
+      },
+      "Lizard": {
+        "allOf": [
+          {
+            "$ref": "#/components/schemas/Pet"
+          },
+          {
+            "type": "object",
+            "properties": {
+              "lovesRocks": {
+                "type": "boolean"
+              }
+            }
+          }
+        ]
       }
     }
   }
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -373,17 +373,40 @@ export default class ApiGenerator {
     onlyMode?: OnlyMode,
     ignoreDiscriminator?: boolean,
   ) {
-    let { $ref } = obj;
-    if (ignoreDiscriminator) $ref = "ignoreDiscriminator:" + $ref;
+    const available = ($ref: string) => {
+      try {
+        this.resolve({ $ref });
+        return false;
+      } catch (error) {
+        return true;
+      }
+    };
+
+    const $ref = ((
+      obj: OpenAPIV3.ReferenceObject,
+      ignoreDiscriminator: boolean,
+    ) => {
+      let { $ref } = obj;
+      if (!ignoreDiscriminator) return $ref;
+
+      $ref += "Base";
+      if (available($ref)) return $ref;
+
+      let i = 2;
+      while (true) {
+        const $ref_ = $ref + String(i);
+        if (available($ref_)) return $ref_;
+        i += 1;
+      }
+    })(obj, ignoreDiscriminator ?? false);
+
     if (!this.refs[$ref]) {
       let schema = this.resolve<SchemaObject>(obj);
       if (ignoreDiscriminator) {
         schema = _.cloneDeep(schema);
         delete schema.discriminator;
       }
-      const name =
-        (schema.title || getRefName($ref)) +
-        (ignoreDiscriminator ? "Base" : "");
+      const name = schema.title || getRefName($ref);
       const identifier = toIdentifier(name, true);
 
       // When this is a true enum we can reference it directly,
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -56,7 +56,7 @@ type SchemaObject = OpenAPIV3.SchemaObject & {
   "x-enumNames"?: string[];
   "x-enum-varnames"?: string[];
   prefixItems?: (OpenAPIV3.ReferenceObject | SchemaObject)[];
-  metadata?: SchemaMetadata; // information oazapfts uses while generating code
+  metadata?: SchemaMetadata; // information added by oazapfts
 };
 
 /**
@@ -288,8 +288,7 @@ export default class ApiGenerator {
     public readonly isConverted = false,
   ) {}
 
-  // Store which schemas are parent schemas.
-  // See `preprocessSchemas` for the definition of a parent schema.
+  // Store parent schemas (see `preprocessSchemas` for the definition of a parent schema)
   parentSchemaRefs: Set<string> = new Set();
 
   aliases: (ts.TypeAliasDeclaration | ts.InterfaceDeclaration)[] = [];
@@ -968,11 +967,10 @@ export default class ApiGenerator {
   /**
    * Does three things:
    * 1. Add a `metadata` property.
-   * 2. Record which schemas are parent schemas in `this.parentSchemaRefs`. The parent schema refers
-   *    to the schema that has a `discriminator` property which is neither used in conjunction with
-   *    `oneOf` nor `anyOf`.
-   * 3. Make all mappings of parent schemas explicit in order to generate types of parent schemas on
-   *    the spot.
+   * 2. Record parent schemas in `this.parentSchemaRefs`. A parent schema refers to a schema that
+   *    has a `discriminator` property which is neither used in conjunction with `oneOf` nor
+   *    `anyOf`.
+   * 3. Make all mappings of parent schemas explicit to generate types immediately.
    */
   preprocessSchemas(schemas: {
     [key: string]: OpenAPIV3.ReferenceObject | SchemaObject;
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -45,18 +45,14 @@ export function getBodyFormatter(body?: OpenAPIV3.RequestBodyObject) {
   }
 }
 
-type SchemaMetadata = {
-  selfRef?: string;
-};
-
 // Augment SchemaObject type to allow slowly adopting new OAS3.1+ features
 // and support custom vendor extensions.
 type SchemaObject = OpenAPIV3.SchemaObject & {
   const?: unknown;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by Xiphe** [comment]
Hi @bdm-k Thanks for the PR! 
Just to let you know: I'm in the middle of reviewing this, but need to do some other things before I can finish it.

File: src/codegen/generate.ts [hunk 1]
```diff
@@ -908,6 +964,56 @@ export default class ApiGenerator {
...
+        ] as SchemaObject;
+        const discriminator = parentSchema.discriminator!;
+        if (isExplicit(discriminator, prefix + name)) continue;
+        if (!discriminator.mapping) {
+          discriminator.mapping = {};
+        }
+        discriminator.mapping[name] = prefix + name;
+      }
+    }
+  }
```
Xiphe, line 1015: I find this part very hard to scan and to reason about.

Would it be possible to maybe break it up a bit and/or add more comments?

src/codegen/generate.ts
```diff
@@ -908,6 +964,56 @@ export default class ApiGenerator {
...
 
+  /**
+   * Does three things:
+   * 1. Add a `metadata` property.
+   * 2. Record parent schemas in `this.parentSchemaRefs`. A parent schema refers to a schema that
+   *    has a `discriminator` property which is neither used in conjunction with `oneOf` nor
+   *    `anyOf`.
+   * 3. Make all mappings of parent schemas explicit to generate types immediately.
+   */
+  preprocessSchemas(schemas: {
```
Xiphe, line 975: Can we rename this to `preprocessComponents`?

src/codegen/generate.ts
```diff
@@ -45,13 +45,18 @@ export function getBodyFormatter(body?: OpenAPIV3.RequestBodyObject) {
   }
 }
 
+type SchemaMetadata = {
+  selfRef?: string;
+};
```
Xiphe, line 50: Can we use a more single-purpose structure for this? 
`metaData.selfRef` Isn't really telling. 

How about `x-component-ref-path`?

src/codegen/generate.ts
```diff
@@ -908,6 +964,56 @@ export default class ApiGenerator {
...
+  preprocessSchemas(schemas: {
+    [key: string]: OpenAPIV3.ReferenceObject | SchemaObject;
+  }) {
+    const prefix = "#/components/schemas/";
+
+    for (const name of Object.keys(schemas)) {
+      const schema = schemas[name];
+      if (isReference(schema)) continue;
+      schema.metadata = { selfRef: prefix + name };
+      if (!schema.discriminator || !!schema.oneOf || !!schema.anyOf) continue;
```
Xiphe, line 984: Nit: Would prefer `Boolean(schema.oneOf)` instead of `!!schema.oneOf`

src/codegen/generate.ts
```diff
@@ -283,6 +288,9 @@ export default class ApiGenerator {
     public readonly isConverted = false,
   ) {}
 
+  // Store parent schemas (see `preprocessSchemas` for the definition of a parent schema)
+  parentSchemaRefs: Set<string> = new Set();
```
Xiphe, line 292: Can we call this `anyOfOneOfComponents`?

src/codegen/generate.ts
```diff
@@ -360,12 +368,26 @@ export default class ApiGenerator {
...
    */
-  getRefAlias(obj: OpenAPIV3.ReferenceObject, onlyMode?: OnlyMode) {
-    const { $ref } = obj;
+  getRefAlias(
+    obj: OpenAPIV3.ReferenceObject,
+    onlyMode?: OnlyMode,
+    ignoreDiscriminator?: boolean,
+  ) {
+    let { $ref } = obj;
+    if (ignoreDiscriminator) $ref = "ignoreDiscriminator:" + $ref;
```
Xiphe, line 381: Could you explain your thinking behind the `"ignoreDiscriminator"` prefix? Where is it being picked up?

Xiphe (changes requested):
Thanks a lot for the contribution! 

I'm not sure if I like the balance between how niche the requirement is vs. how much more complexity it adds to the implementation. Especially the need to preProcess all components. 
But I also don't think theres an obvious easier way.

Would like to get @fgnass and/or @jonkoops thoughts on this.

Left some comments in regards to naming and code-understandability.

src/codegen/__fixtures__/allOf.json
```diff
@@ -74,6 +105,66 @@
...
+        "properties": {
+          "petType": {
+            "type": "string"
+          }
+        },
+        "required": ["petType"],
+        "discriminator": {
+          "propertyName": "petType",
+          "mapping": {
+            "dog": "#/components/schemas/Dog"
```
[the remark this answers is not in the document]
bdm-k: Yes, you're right. This example is from [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#discriminator-object).

src/codegen/generate.ts
bdm-k: Take the test I added as an example. We need to invent a unique reference for `PetBase` to take advantage of `refs` field. Plus, the reference must be different from "#/components/schemas/Pet". I could use "#/components/schemas/PetBase", but there might already be a schema named "PetBase". So, I've decided to use "ignoreDiscriminator:#/components/schemas/Pet".

src/codegen/generate.ts
bdm-k: I agree that "parent schema" is too general and I'm open to using a more suitable term if there is one. However, I feel `anyOfOneOfComponents` is kind of confusing. Shouldn't the variable represent components that have a `discriminator` but don't have `oneOf` or `anyOf`?

src/codegen/generate.ts
bdm-k: > Will we have both refs `ignoreDiscriminator:#/components/schemas/Pet` and `#/components/schemas/Pet` on the list?

Yes.

bdm-k: Okay, I'll think about this. If you come up with a specific implementation, please send a PR to my branch anytime. I'll merge it :)

src/codegen/generate.ts
bdm-k: Looking at other field variables, "refs" seems to be used to mean `ts.TypeReferenceNode`. Therefore, the name `parentSchemaRefs` could be a bit misleading. I'm going to change it to `parentSchemas`. As for "parent schema", I haven't come up with a better idea yet.

src/codegen/generate.ts
Xiphe: Would love to get more into this but unfortunately need to balance my time in favor of other projects currently.

I was thinking that maybe this also is just a naming/understanding/clarity problem. 
A common pattern I notice in my head when reviewing is not understanding how all the parts are supposed to play together. 
Maybe we can up with a good name for the pattern _(unfortunately the Spec also just talks about the HOW, and not about the WHAT)_ and then align our internal naming & comments to that?

But again: I notice that I'm not really constructive here and I'm grateful for the work you put in, please take my comments as inspiration rather then requirements. We have tests, other tests aren't failing and that also means we can as well refactor the internals later once we learned more about it or run into actual issues.

Xiphe: Another Thought: What if we actually go with `#/components/schemas/PetBase` and do a check beforehand if that name is actually free? (And if it isn't go with `#/components/schemas/PetBase2`?)

bdm-k: > What if we actually go with `#/components/schemas/PetBase` and do a check beforehand if that name is actually free? (And if it isn't go with `#/components/schemas/PetBase2`?)

That sounds natural to me. I'll give it a try.

Xiphe:
Please let me know when this is ready for another review or need additional input. 🤙

bdm-k:
I'm ready for another review.

src/codegen/generate.ts
```diff
@@ -360,11 +364,48 @@ export default class ApiGenerator {
...
+      $ref += "Base";
+      if (available($ref)) return $ref;
+
+      let i = 2;
+      while (true) {
+        const $ref_ = $ref + String(i);
+        if (available($ref_)) return $ref_;
+        i += 1;
+      }
+    })(obj, ignoreDiscriminator ?? false);
```
Xiphe, line 401: Can we move this into a standalone function? like `findAvailableRefName`?

src/codegen/generate.ts
```diff
@@ -360,11 +364,48 @@ export default class ApiGenerator {
 
   /**
    * Create a type alias for the schema referenced by the given ReferenceObject
+   * @param ignoreDiscriminator If true, the discriminator property of the schema referenced by
```
Xiphe, line 367: When I see `@param` I expect [JSDoc](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) format. Would propose to move the comment inline just above `ignoreDiscriminator?: boolean,`

src/codegen/generate.ts
```diff
@@ -539,11 +580,52 @@ export default class ApiGenerator {
...
+      // parent schema -> union
+      const mapping = schema.discriminator.mapping || {};
+      return this.getUnionType(
+        Object.values(mapping).map((ref) => ({ $ref: ref })),
+        undefined,
+        onlyMode,
+      );
+    }
     if (schema.allOf) {
       // allOf -> intersection
```
Xiphe, line 593: I think this comment should move down along with the old implementation of this, right?

src/codegen/generate.ts
```diff
@@ -283,6 +284,9 @@ export default class ApiGenerator {
     public readonly isConverted = false,
   ) {}
 
+  // Store parent schemas (see `preprocessComponents` for the definition of a parent schema)
+  parentSchemas: Set<string> = new Set();
```
Xiphe, line 288: Would still like to give this one a name that's more clear towards why this exists. I feel like `parent` is too generic.

Can we use `discriminatingSchemas` maybe?

Xiphe (changes requested):
This is going into a good direction! Just a few more naming/documenting things and we're good to go i think.

src/codegen/generate.ts
bdm-k: OK. Maybe we should name the return value of `findAvailableRefName` something other than `$ref`, since it's no longer the `$ref` property of the `obj`. How about simply using `key`?

src/codegen/generate.ts
bdm-k: Where exactly should the comment be? An intersection type is still generated in the new implementation, so I think it's ok to leave the comment in its current position.

src/codegen/generate.ts
bdm-k: nice name!

src/codegen/generate.ts
Xiphe: yeah, sounds feasible to me 👍

src/codegen/generate.ts
Xiphe: ok, let's leave it here then

src/codegen/generate.ts
bdm-k: I've decided not to change the name after all.

Xiphe (approved):
LGTM 🚀

Xiphe:
Awesome! Thanks a bunch for adding this in @bdm-k 🫶 Can we merge?

bdm-k:
Absolutely! Thanks for the feedback. Let's merge it. 👍

PR #479, opened 2023-09-12
Fix multipart uploading

Xiphe:
Up until now we had no way of handling 

1. A previously unknown amount of file uploads
2. File uploads along with complex data

This takes inspiration from https://github.com/oazapfts/oazapfts/pull/472


Xiphe:
pre-released as `oazapfts@4.9.0-fix-multipart-uploading.1`

PR #489, opened 2023-10-03
Prevent infinite recursion of checkSchemaOnlyMode

### LINKED ISSUE #486: Recursive $ref in components
Hi,

I have an endpoint that returns a list of folders and subfolders. For that the definition of the schemas is recursive like this:

`            "FolderDto": {
                "type": "object",
                "properties": {
                    "name": {
                        "type": "string"
                    },
                    "files": {
                        "type": "array",
                        "items": {
                            "type": "string"
                        }
                    },
                    "folders": {
                        "type": "array",
                        "items": {
                            "$ref": "#/components/schemas/FolderDto"
                        }
                    }
                },
                "required": []
            }`

This self reference gives an error

`node_modules/oazapfts/lib/codegen/generate.js:136
        .map((s) => decodeURI(s.replace(/~1/g, "/").replace(/~0/g, "~")));
             ^

RangeError: Maximum call stack size exceeded`

As soon as I remove the $ref it's working fine.

Thanks for looking!

bdm-k:
Closes #486 

In the current implementation, `checkSchemaOnlyMode` does not work for recursive schemas (i.e. encounter inifinite recursion). This PR addresses the problem.

### DIFF
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -776,72 +776,46 @@ export default class ApiGenerator {
       return false;
     }
 
-    return this.checkSchemaOnlyModeHelper(
-      schema,
-      onlyMode,
-      resolveRefs,
-      new Set<string>(),
-    );
-  }
+    const check = (
+      schema: SchemaObject | OpenAPIV3.ReferenceObject,
+      history: Set<string>,
+    ): boolean => {
+      if (isReference(schema)) {
+        if (!resolveRefs) return false;
 
-  checkSchemaOnlyModeHelper(
-    schema: SchemaObject | OpenAPIV3.ReferenceObject,
-    onlyMode: OnlyMode,
-    resolveRefs: boolean,
-    history: Set<string>,
-  ): boolean {
-    if (isReference(schema)) {
-      if (!resolveRefs) return false;
+        // history is used to prevent infinite recursion
+        if (history.has(schema.$ref)) return false;
 
-      // history is used to prevent infinite recursion
-      if (history.has(schema.$ref)) return false;
+        history.add(schema.$ref);
+        const ret = check(this.resolve(schema), history);
+        history.delete(schema.$ref);
+        return ret;
+      }
 
-      history.add(schema.$ref);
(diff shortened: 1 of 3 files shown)


src/codegen/generate.ts
```diff
@@ -776,10 +776,35 @@ export default class ApiGenerator {
...
 
+    return this.checkSchemaOnlyModeHelper(
+      schema,
+      onlyMode,
+      resolveRefs,
+      new Set<string>(),
+    );
+  }
+
+  checkSchemaOnlyModeHelper(
```
Xiphe, line 787: Do you think it's possible to merge `checkSchemaOnlyModeHelper` and `checkSchemaOnlyMode`? Or otherwise use a more clear name instead of the helper? Looks like something that should be internal API...

Xiphe (commented):
Awesome! Thanks for working on this. As usual can we also add a test-case to ensure that we don't run into the same problem again in case we ever refactor the generate class?

src/codegen/generate.ts
bdm-k: We can define it as a nested function in `checkSchemaOnlyMode`. How do you think of this solution?

Xiphe: Would prefer that, yeah

src/codegen/index.test.ts
```diff
@@ -69,6 +69,13 @@ describe("generateSource", () => {
     );
   });
 
+  it("should support recursive schemas", async () => {
+    const src = await generate("/__fixtures__/recursive.yaml");
+    expect(src).toContain(
+      "export type FolderDto = { name?: string; files?: string[]; folders?: FolderDto[]; };",
+    );
+  });
+
```
Xiphe, line 78: 💯👍

mikonse:
Are there any news on this?
Having a release with this change integrated would really help with a project since I'm currently stuck with manually patching the oazapts sources in the node_modules

bdm-k:
I heard that @Xiphe would be offline until the end of November. This PR will be merged after that.

noneall:
hi, any updates on this?

Issue #491, opened 2023-10-10
4.9.0 and 4.10.0 generate circular types


phryneas:
I just noticed that while updating dependencies for the RTK Query Codegen - 4.9.0 and 4.10.0 seem to generate this impossible type:

![grafik](https://github.com/oazapfts/oazapfts/assets/4282439/8a6f0e3d-daf0-4f68-b69e-146099e00a6e)

You can find the schema for that here: https://github.com/reduxjs/redux-toolkit/blob/6a24024b67f847fd06c24d82b95dda3a8e648359/packages/rtk-query-codegen-openapi/test/fixtures/issue-2002.json

bdm-k:
Hello, @phryneas  
Thank you for reporting. It seems to be related to the commit (#473) I made.

I generated code from the OpenAPI Document at the link above, using v4.10.0.
```typescript
// I formatted the code a bit for readability
export type AnimalType = "All" | "Cats" | "Dogs";
export type AnimalBase = {
    "type": AnimalType;
    id?: number;
    name?: string | null;
};
export type Dog = { "type": "Dog" } & AnimalBase & {
    dogUniqueProp?: string | null;
};
export type Cat = { "type": "Cat" } & AnimalBase & {
    catUniqueProp?: string | null;
};
```
The result was different from the one shown in your image.

Could you please provide more information about your environment?

phryneas:
This is one of our user-provided unit tests, so I'll try to sum it up as good as possible ^^

You can see the whole thing happening here:
https://github.com/reduxjs/redux-toolkit/pull/3791

The underlying test is this one (sorry it doesn't show a lot):
https://github.com/reduxjs/redux-toolkit/blob/067b6d40a681eaa269a0d4e1511e17d4359ad2e5/packages/rtk-query-codegen-openapi/test/generateEndpoints.test.ts#L344-L351

Oazapfts itself is created here and called throughout that file:
https://github.com/reduxjs/redux-toolkit/blob/067b6d40a681eaa269a0d4e1511e17d4359ad2e5/packages/rtk-query-codegen-openapi/src/generate.ts#L101-L104

I have to admit that I haven't touched the Codegen in a while (I'm absolutely swamped with Open Source) and we've just merged a bunch of user PRs, so I'm not 100% up to speed what's actually going on in there 😞 
I hope this is already enough for you to go on.

bdm-k:
[...]

PR #525, opened 2023-11-27
Do preprocessing in the constructor

bdm-k:
The following code has been moved from `generateApi()` to `constructor()`.
```typescript
if (this.spec.components?.schemas) {
  this.preprocessComponents(this.spec.components.schemas);
}
```
The motivation behind this change is described in #491.


Xiphe (approved):
Makes sense! Thank you

LGTM :gorilla:

PR #537, opened 2023-12-12
Let Fetch API set the Content-Type header

### LINKED ISSUE #512: Multipart upload missing the boundary header
Hi,

when I generate a multipart upload I can't post it to my server due to the missing header "boundary". The header that gets generated is just `'Content-Type': 'multipart/form-data'`. Using for example Insomnia to generate the same request gives something like `'Content-Type': 'multipart/form-data; boundary=---011000010111000001101001'`. When using fetch directly the same behavior occurs. With fetch you have to explicitly not set the header and let the browser generate it for you. In my case this does not work:

```const form = new FormData();
  form.append("file", file);

  const options_ = {
    method: "PUT",
    body: form,
    headers: { "Content-Type": multipart/form-data },
  };

  return fetch("/api/v1/files/", options_)
```
but this does

```
const form = new FormData();
  form.append("file", file);

  const options_ = {
    method: "PUT",
    body: form,
  };

  return fetch("/api/v1/files/", options_)
```

In the second case the header is generated by the browser. Also see [here](https://muffinman.io/blog/uploading-files-using-fetch-multipart-form-data/) or [here](https://stackoverflow.com/questions/39280438/fetch-missing-boundary-in-multipart-form-data-post). Some time ago this was working fine.

Thank you for your time!

bdm-k:
Closes #512 

This RP has removed the Content-Type header from multipart/form-data requests. This is because the Content-Type header should be automatically set by the Fetch API.

### DIFF
--- a/src/runtime/index.ts
+++ b/src/runtime/index.ts
@@ -114,12 +114,6 @@ export function runtime(defaults: RequestOpts) {
       : "application/x-www-form-urlencoded";
   }
 
-  function ensureMultipartContentType(contentTypeHeader: string) {
-    return contentTypeHeader?.startsWith("multipart/form-data")
-      ? contentTypeHeader
-      : "multipart/form-data";
-  }
-
   return {
     ok,
     fetchText,
@@ -177,15 +171,14 @@ export function runtime(defaults: RequestOpts) {
         }
       });
 
+      // Remove the Content-Type header because it should be automatically set
+      // by the Fetch API.
+      delete headers?.["Content-Type"];
+
       return {
         ...req,
         body: data,
-        headers: {
-          ...headers,
-          "Content-Type": ensureMultipartContentType(
-            String(headers?.["Content-Type"]),
-          ),
-        },
+        headers,
       };
     },
   };
--- a/demo/api.test.ts
+++ b/demo/api.test.ts
@@ -387,10 +387,7 @@ describe("multipart", () => {
       "http://localhost:8000/v2/pet/5/uploadImage",
       expect.objectContaining({
         body: expect.any(FormData),
-        headers: {
-          Accept: "application/json",
-          "Content-Type": "multipart/form-data",
-        },
+        headers: { Accept: "application/json" },
         method: "POST",
       }),
     );


src/runtime/index.ts
```diff
@@ -177,15 +171,14 @@ export function runtime(defaults: RequestOpts) {
         }
       });
 
+      // Remove the Content-Type header because it should be automatically set
+      // by the Fetch API.
+      delete headers?.["Content-Type"];
```
Xiphe, line 176: I think will also delete "Content-Type" headers intentionally set by the user. Which is behavior I wouldn't expect.

Xiphe (changes requested):
I think this approach isn't ideal.

When I understand https://github.com/oazapfts/oazapfts/issues/512#issuecomment-1829480876 correctly we'd only need to NOT set the default headers. 

When the users actively sets a custom content-type we must not `delete` it. That's a [feature](https://github.com/oazapfts/oazapfts/pull/456).

We intended to [remove the default content-type headers](https://github.com/oazapfts/oazapfts/blob/b18de774725da16d9e7b68d684a3c4271ff4cedf/src/runtime/index.ts#L103C2-L121C4) in a future breaking change. So if we can double-check that just removing the defaults would also fix https://github.com/oazapfts/oazapfts/issues/512 we could move forward with releasing a v5.

what do you think?

src/runtime/index.ts
bdm-k: My understanding is that when the request body is `FormData`, you must not specify anything in the Content-Type header. Therefore, if the Content-Type header is specified at this point, it would be a error on the user's side.
That said, I don't think we need to go out of our way to correct such mistakes, so I'll follow your approach.

bdm-k:
I agree with removing the default Content-Type headers. I've just confirmed that this fixes #512, but it has cause some test cases to fail. I'm updating these test cases now.

bdm-k:
@Xiphe I just want to make sure, by removing the default Content-Type headers, do you mean something like this?
```diff
headers: {
  ...headers,
- "Content-Type": ensureJsonContentType(
-   String(headers?.["Content-Type"]),
- ),
+ "Content-Type": headers?.["Content-Type"] ?? "application/json",
}
```

As for versions, it might be a good idea to release this PR as is, and then release the version 5 in another PR. After all, it is not good to leave the version 4 buggy.

Xiphe:
Yeah that looks about right. I'd suggest to do also allow sending no `Content-Type` header by for example setting it to `null` or a `Symbol('no content type')`.

As to your suggestion to merge this as is: 

From what I understand multi-part uploads were not functional before https://github.com/oazapfts/oazapfts/commit/b0058dee6183256bbff6c4f0562a8775305a0c11 which introduced this behavior as is. Which means either, the people reporting on https://github.com/oazapfts/oazapfts/issues/512 are the only ones using multi-part uploading and the functionality is indeed still not operational or the issue describes an edge-case. When the latter is the case, then the fix in this PR might break behavior for users that have made multi-part uploads work with the current v4. 

On a general level I think reliable change communication is more important then having every version free of bugs. 
When v4 has a bug that we can not fix in a backwards compatible way we release v5.

Xiphe:
Gonna build a v5 now and we can discuss wether we want to fix v4 then :)

Xiphe:
=> https://github.com/oazapfts/oazapfts/pull/538

bdm-k:
Makes sense! Let's get started on version 5!

Issue #509, opened 2023-11-03
Exclude oazapt’s default headers from generated function parameters


hpohlmeyer:
The docs recommend setting defaults for commonly used parameters.
Unfortunately they are still included in the generated functions, when they are specified in the spec.

Here is an example schema:
```json
{
  "paths": {
    "/user/{id}": {
      "get": {
        "parameters": [
          {
            "schema": {
              "type": "string"
            },
            "in": "path",
            "name": "id",
            "required": true
          },
          {
            "schema": {
              "type": "string"
            },
            "in": "header",
            "name": "x-some-info",
            "required": true
          }
        ],
        // …
      }
    }
  }
}
```

The `x-some-info` header is required on the request, which means it should also appear in the schema.
But since I want it on almost every request, I would add it to the oazapfts defaults:

```js
import * as api from "./api.ts";

api.defaults.headers = {
  "x-some-info": "my-info",
};
```

When I use oazapfts now to generate my api, the resulting function looks like this:

```ts
function getUserById(id: string, xSomeInfo: string, opts?: Oazapfts.RequestOpts) {
  // ...
}
```

This forces me to add `xSomeInfo` every time, even if I know it will be supplied by the defaults.
It would be nice to pass a list of headers to oazapfts that should be ignored during code generation.

I am thinking of something like

```sh
oazapfts --default-headers=x-some-info,x-other-info
```

or 

```sh
oazapfts --default-header=x-some-info --default-header=x-other-info
```

hpohlmeyer:
I am currently using `patch-package` to accomplish that and would be happy to open a PR if the flag names are fine with you.

Xiphe:
Hmm. I understand the problem, thanks for bringing this up!

That said, I'm not entirely sure if it's worth extending the api of this lib for this.
For your case, the `defaultHeaders` configuration would be sufficient but extending on that thought one might want to filter some other request parameters, too. Maybe also remove some parameters only on specific endpoints. And I don't see us going into that direction.

Without digging really deep into this I'd initially argue that when a header is valid and constant for each method of an API it should not be part of the endpoint-spec but instead part of a global configuration. Meaning, removing the parameter from the spec. 

Even for externally provided specs I'd consider adding a "we know better"-pre-processing of the spec before feeding it into oazapfts a pattern I'd favor here.

Let me know what you think

PR #538, opened 2023-12-14
Smarter header defaults and handling

Xiphe:
- stop setting Content-Type header for multipart requests
- support HeaderInit in request objects
- support default headers

BREAKING CHANGE:
multipart/form-data headers are not set by oazapfts anymore as these are usually automatically set by the browser/node when this causes problems on your end you can set Content-Type=multipart/form-data manually via RequestOpts

oazapfts runtime now works with `Header` instead of plain Objects `{}`. This might affect you when you use a custom fetch implementation and manipulate headers there

oazapfts no longer safe-guards "Content-Type" headers
in the past we discarded custom Content-Type headers that didn't start with `json` for json requests
and similarly Content-Type headers that didn't start with application/x-www-form-urlencoded for form requests
we now trust you to set the correct headers for your requests

--

fix https://github.com/oazapfts/oazapfts/issues/512
ref https://github.com/oazapfts/oazapfts/issues/509

### DIFF
--- a/demo/api.test.ts
+++ b/demo/api.test.ts
@@ -387,13 +387,12 @@ describe("multipart", () => {
       "http://localhost:8000/v2/pet/5/uploadImage",
       expect.objectContaining({
         body: expect.any(FormData),
-        headers: {
-          Accept: "application/json",
-          "Content-Type": "multipart/form-data",
-        },
         method: "POST",
       }),
     );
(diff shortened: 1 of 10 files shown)


Xiphe:
pre-released as `oazapfts@5.0.0-headers.1`

PR #540, opened 2023-12-18
feat: Include property descriptions

Duckers:
This emits `SchemaObject.description` in a multi-line comment in front of the property.

This makes sure useful documentation is included in the generated code.

### DIFF
--- a/demo/api.ts
+++ b/demo/api.ts
@@ -28,9 +28,13 @@ export type Pet = {
   name: string;
   photoUrls: string[];
   tags?: Tag[];
+  /** pet status in the store */
   status?: "available" | "pending" | "sold" | "private" | "10percent";
+  /** Always true for a pet */
   animal?: true;
+  /** Size scale for pets */
   size?: "P" | "M" | "G" | "0";
+  /** integer test case for #349 */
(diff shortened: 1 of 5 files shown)


Xiphe:
Awesome! Would you mind running `npm run generate-demo` and check in the changes to the demo files? Or are there no descriptions in the demo files?

Duckers:
@Xiphe 
image

demo/api.ts
```diff
@@ -28,9 +28,13 @@ export type Pet = {
   name: string;
   photoUrls: string[];
   tags?: Tag[];
+  /*pet status in the store*/
```
Xiphe, line 31: Can we make those JSDoc Comments like `/** pet status in the store */`?
That way the description will actually be picked up by editors (VSCode at least does not show normal comments, only JSDoc).

image

image

Xiphe (changes requested):

Xiphe:
Thank you for the update! I think making those JSDoc comments would be even more helpful.

demo/api.ts
Duckers: Done

Xiphe (approved):

Xiphe:
Thank you for the contribution!

Issue #545, opened 2023-12-28
Comment escape sequence in OpenAPI description needs to be escaped.


josh-:
I'm using oazapfts with the [Jira Cloud REST API](https://dac-static.atlassian.com/cloud/jira/platform/swagger-v3.v3.json?_v=1.6768.0-0.1297.0) which includes this in `TimeTrackingProvider`:

```json
"url": {
    "description": "The URL of the configuration page for the time tracking provider app. For example, */example/config/url*. This property is only returned if the `adminPageKey` property is set in the module descriptor of the time tracking provider app.",
    "readOnly": true,
    "type": "string"
}
```

Since this description includes a `*/` value, the resulting TypeScript code is not valid as the comment ends mid-way through the string:

```ts
    /** The URL of the configuration page for the time tracking provider app. For example, */example/config/url*. This property is only returned if the `adminPageKey` property is set in the module descriptor of the time tracking provider app. */
    url?: string;
```

```
SyntaxError: Property or signature expected. (811:94)
```

The `*/` in descriptions should be escaped before being added to the TS comment

PR #546, opened 2023-12-30
Escape any comment closing tags in schema descriptions

josh-:
Fixes https://github.com/oazapfts/oazapfts/issues/545.

Before:

```ts
/** The URL of the configuration page for the time tracking provider app. For example, */example/config/url*. This property is only returned if the `adminPageKey` property is set in the module descriptor of the time tracking provider app. */
url?: string;
```

After:

```ts
/** The URL of the configuration page for the time tracking provider app. For example, *\/example/config/url*. This property is only returned if the `adminPageKey` property is set in the module descriptor of the time tracking provider app. */
url?: string;
```

### DIFF
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -903,11 +903,14 @@ export default class ApiGenerator {
       });
 
       if ("description" in schema && schema.description) {
+        // Escape any JSDoc comment closing tags in description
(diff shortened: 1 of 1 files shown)


Xiphe:
Thanks for the contribution!

Issue #541, opened 2023-12-21
ESM Typescript / ERR_UNSUPPORTED_DIR_IMPORT


florianbepunkt:
Using this lib in a NodeJS backend within an ESM module, I get the following error: `ERR_UNSUPPORTED_DIR_IMPORT`

I need to manually change
```ts
// from 
import * as Oazapfts from "oazapfts/lib/runtime";
import * as QS from "oazapfts/lib/runtime/query";

// to 
import * as Oazapfts from "oazapfts/lib/runtime/index.js";
import * as QS from "oazapfts/lib/runtime/query.js";
```

jonkoops:
This is how ESM works, this is not a bug in Oazapfts. If you import a file in ESM you will have to specify the full path to it. That said, you should not import files from anything but the top level `oazapfts` namespace. Importing things from `/lib/` will likely break in the future.

florianbepunkt:
Thank you for explaining how ESM works. 

So you need to fix the auto generated code by hand if you use ESM in a Node / Typescript setup. Is this intended? Looks like a bug to me, if you need to fix the module resolution of auto-generated code.

if the lib is not compatible with Node ESM/Typescript setup (which I can understand), a warning in the docs would be nice. I can open a PR if you like.

jonkoops:
Ah, my bad, I did not realize this code was generated. Let me re-open this.

jonkoops:
@Xiphe we'll need to use the [`exports` field](https://nodejs.org/api/packages.html#package-entry-points) in the `package.json` and add explicit entry-points for `oazapfts/runtime` and `oazapfts/codegen` which exports all the required API surface.

Xiphe:
Yeah, that makes sense. For documentation and reproducibility I would like to understand how the generated ts file was transpiled to js

florianbepunkt:
> You're right that the auto-generated code of this lib should support all sorts of environments.
> 
> So to clarify: You're referring to line 7+8 of an API client that has been generated with oazapfts, right? 

Yes.

> As the artifact is a `.ts` file I had expect that the typescript compiler would handle the conversation between different module systems?

Unfortunately not, there is currently an experimental flag to resolve extensionless imports, but afaik this is node's esm behaviour (not a typescript issue. You need to specify full import paths, including file extension (the compiled one, so .js). Also for directory imports with barrel files, you need to specify the barrel file.

So:

```ts
// this... 
import * as Oazapfts from "oazapfts/lib/runtime";
[...]

PR #550, opened 2024-01-10
fix: add export fields for use in esm

Xiphe:
fix https://github.com/oazapfts/oazapfts/issues/541

pre-released as `oazapfts@5.0.3-exports.2`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,38 @@
   "version": "0.0.0-development",
   "description": "OpenApi TypeScript client generator",
   "main": "lib/index.js",
+  "exports": {
+    "./lib/runtime": {
+      "require": "./lib/runtime/index.js",
+      "default": "./lib/runtime/index.js",
+      "types": "./lib/runtime/index.d.ts"
+    },
+    "./lib/runtime/query": {
+      "require": "./lib/runtime/query.js",
+      "default": "./lib/runtime/query.js",
+      "types": "./lib/runtime/query.d.js"
+    },
+    "./lib/runtime/index.js": {
+      "require": "./lib/runtime/index.js",
+      "default": "./lib/runtime/index.js",
+      "types": "./lib/runtime/index.d.ts"
+    },
+    "./lib/runtime/query.js": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,6 +3,18 @@
   "version": "0.0.0-development",
   "description": "OpenApi TypeScript client generator",
   "main": "lib/index.js",
+  "exports": {
```
jonkoops, line 6: Note that this restricts all previous exports as well, so the following will no longer work:

```ts
import * as Oazapfts from "oazapfts/lib/runtime/index.js";
import * as QS from "oazapfts/lib/runtime/query.js";
```

So this might be considered a breaking change, unless all previously existing paths are added to the `export` config as well.

package.json
```diff
@@ -3,6 +3,18 @@
...
+    "./lib/runtime": {
+      "require": "./lib/runtime/index.js",
+      "default": "./lib/runtime/index.js",
+      "types": "./lib/runtime/index.d.ts"
+    },
+    "./lib/runtime/query": {
+      "require": "./lib/runtime/query.js",
+      "default": "./lib/runtime/query.js",
+      "types": "./lib/runtime/query.d.js"
+    }
```
jonkoops, line 16: Perhaps this can now be simplified so that the imports are always done from the root:

```ts
import * as Oazapfts from "oazapfts/runtime";
import * as QS from "oazapfts/runtime/query";
```

```suggestion
    "./runtime": {
      "require": "./lib/runtime/index.js",
      "default": "./lib/runtime/index.js",
      "types": "./lib/runtime/index.d.ts"
    },
    "./runtime/query": {
      "require": "./lib/runtime/query.js",
      "default": "./lib/runtime/query.js",
      "types": "./lib/runtime/query.d.js"
    }
```

package.json
Xiphe: Good point!

package.json
Xiphe: I like this. Will only prepare the exports for this though and add a note for next major. This might break in old environments that do not respect the `exports` field

PR #556, opened 2024-01-19
fix: move default exports to bottom

Thijsdh:
The changes made in https://github.com/oazapfts/oazapfts/commit/768ddd404ea256a46b92af949e444133119c5bd6 broke my build with the following webpack error after running `next build`:

```ts
Module not found: Default condition should be last one
   6 |  * See https://www.npmjs.com/package/oazapfts
   7 |  */
>  8 | import * as Oazapfts from "oazapfts/lib/runtime";
   9 | import * as QS from "oazapfts/lib/runtime/query";
  10 | export const defaults: Oazapfts.Defaults<Oazapfts.CustomHeaders> = {
  11 |     headers: {},

https://nextjs.org/docs/messages/module-not-found
```

According to the [webpack documentation](https://webpack.js.org/guides/package-exports/#conditional-syntax):
> The last condition in the object might be the special "default" condition, which is always matched.

Since the `default` condition was not the last one, the error above occurred. This PR moves all `default` conditions to the bottom which should resolve this error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,33 +5,33 @@
   "main": "lib/index.js",
   "exports": {
     "./lib/runtime": {
-      "require": "./lib/runtime/index.js",
       "types": "./lib/runtime/index.d.ts",
+      "require": "./lib/runtime/index.js",
       "default": "./lib/runtime/index.js"
     },
     "./lib/runtime/query": {
-      "require": "./lib/runtime/query.js",
       "types": "./lib/runtime/query.d.js",
+      "require": "./lib/runtime/query.js",
       "default": "./lib/runtime/query.js"
     },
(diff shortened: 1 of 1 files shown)


package.json
jonkoops, line 1: The `types` should actually be the first item in each of these, as per the [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/modules/reference.html#example-explicit-types-condition).

Thijsdh: You're right, I've moved them.

jonkoops (approved):
Chef's kiss

Xiphe:
Thanks for fixing!

Issue #401, opened 2023-03-21
Move typescript from dependencies to devDependencies


Fenntasy:
Hi there,

This may be a problem depending on how the codegen is done but having `typescript` in the `dependencies` of the package mean that have `oazapfts` as a `dependency` for us will add `typescript` as our "production" dependencies.
We're using Docker so our node modules add a whole bunch of data in our resulting container (around 200mb :/).

Is there a possibility of having it switch to dev deps here?

Xiphe:
that's not really an option since, as you've already figured, the codegen part needs it.

The solution would be splitting the package up into a codegen and a runtime part. Then you'd install `oazapfts-gen` (along with typescript) as a dev-dependency in your project  and `oazapfts-runtime` as a runtime dependency. 

This indeed is something I was thinking about for a longer time bug have not found the time to discuss this with the other maintainers and I haven't wrapped my head around it fully in terms of coming up with a migration path.

jonkoops:
We could use NPM workspaces to create two seperate packages and publish them together.

Xiphe:
Would appreciate when someone else could beta-test [oazapfts@6.0.0-beta.1](https://github.com/oazapfts/oazapfts/releases/tag/oazapfts-v6.0.0-beta.1) and [@oazapfts/runtime@1.0.0-beta.1](https://github.com/oazapfts/oazapfts/releases/tag/%40oazapfts%2Fruntime-v1.0.0-beta.1).

Also looking for a review on https://github.com/oazapfts/oazapfts/pull/576. It's a quite extensive changeset though 😬

PR #568, opened 2024-02-06
perf: reduce runtime size by using esnext

benmccann:
This simplifies the generated code a fair amount. E.g. consider the example below:

On `main`:
```js
    function fetchText(url, req) {
        return __awaiter(this, void 0, void 0, function* () {
            const res = yield doFetch(url, req);
            let data;
            try {
                data = yield res.text();
            }
            catch (err) { }
            return {
                status: res.status,
                headers: res.headers,
                contentType: res.headers.get("content-type"),
                data,
            };
        });
    }
```

With this PR:
```js
    async function fetchText(url, req) {
        const res = await doFetch(url, req);
        let data;
        try {
            data = await res.text();
        }
        catch (err) { }
        return {
            status: res.status,
            headers: res.headers,
            contentType: res.headers.get("content-type"),
            data,
        };
    }
```

It's always best for libraries to be distributed as `esnext` as it results in the smallest code. If someone is targeting an older browser or runtime they should be using the appropriate bundler setting to transpile to an older version while building their application rather than expecting each individual library they're using to do so.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,10 +2,10 @@
   "compilerOptions": {
     "outDir": "lib",
     "rootDir": "src",
-    "target": "es2015",
+    "target": "esnext",
     "sourceMap": true,
     "declaration": true,
-    "lib": ["es2015", "dom"],
+    "lib": ["esnext", "dom"],
     "module": "commonjs",
     "strict": true,
     "resolveJsonModule": true,


Xiphe:
Hi @benmccann thank you for the contribution! I fully agree with what you're saying. 

At the same time we should consider this a breaking change as it might require additional tooling in some environments.

I'd therefore either wait with releasing this and bundle it along with the next major release or by doing this backwards compatible.

We could for example start providing and ESM bundle with es2015 around until we'll eventually get rid of it in a major version.

What do you think?

Xiphe:
It's not that complicated actually. Will give it a shot within the next days.

Xiphe:
might actually do split the package into runtime and generator 

ref https://github.com/oazapfts/oazapfts/issues/401

benmccann:
That would be great! Thank you so much

It looks like the `package.json` points at a few files that don't exist, so might be able to clean that up at the same time: https://publint.dev/oazapfts@5.1.4

Xiphe:
Good find. Fixed these directly as I'm a little overwhelmed with other tasks currently and will need to postpone this.

Xiphe:
> image  
>
> **It's not that complicated actually**
> _- Xiphe 2024_


Well i went a little overboard in https://github.com/oazapfts/oazapfts/pull/576. But if you like you could give [oazapfts@6.0.0-beta.1](https://github.com/oazapfts/oazapfts/releases/tag/oazapfts-v6.0.0-beta.1) a try. It exports both, a legacy cjs build targeting es2015 and a esm build targeting esnext.

Going to close here in favor of https://github.com/oazapfts/oazapfts/pull/576.

Huge thanks to your contribution @benmccann  nonetheless as it poked me to do this work that I wanted to do for quite the time now.

benmccann:
Thank you!!

I might suggest that the CJS version should also target esnext, but I'm using ESM myself, so I won't push on that too hard :wink:

Xiphe:
Fair point.

PR #570, opened 2024-02-08
Support type arrays according to OpenAPI spec 3.1.0

### LINKED ISSUE #497: Support `type` as an array
I'm using a downstream library (RTK Query) and have recently updated my spec to 3.1.0.  My spec mostly generates, but it seems multiple values for types (the recommended way to handle 'null' values) is not generating correctly:

```
type:
    - "object"
    - "null" 
```

When run, I get only the 'object' type, and null isn't valid here. Is this something that can be supported?

For more examples I have an issue on RTK here: https://github.com/reduxjs/redux-toolkit/issues/3814, which is where I was redirected to this repository.

Thank you!

SimonEggert:
OpenAPI spec 3.1.0 supports type arrays (s. [Migration Guide](https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0)) and thus becomes [JSON schema compliant](https://json-schema.org/understanding-json-schema/reference/type), e.g.:

```
type:
    - "object"
    - "null" 
```

As far is I understand this is not yet implemented or reflected in this code base. This is an attempt to add the necessary functionality. Please let me know if there are any more changes that need to be made in order to meet the code quality requirements.

closes #497

### DIFF
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -695,7 +695,19 @@ export default class ApiGenerator {
       return this.getTypeFromEnum([schema.const]);
     }
     if (schema.type) {
-      // string, boolean, null, number
+      // string, boolean, null, number, array
+      if (Array.isArray(schema.type)) {
+        return factory.createUnionTypeNode(
+          schema.type.map((type) => {
+            if (type === null) return cg.keywordType.null;
+            if (type === "integer") return cg.keywordType.number;
+            if (type in cg.keywordType)
+              return cg.keywordType[type as keyof typeof cg.keywordType];
+
+            return cg.keywordType.any;
+          }),
+        );
+      }
       if (schema.type === "integer") return cg.keywordType.number;
       if (schema.type in cg.keywordType) return cg.keywordType[schema.type];
     }
--- a/src/codegen/generate.ts
+++ b/src/codegen/generate.ts
@@ -1,7 +1,7 @@
 import _ from "lodash";
 import ts, { factory } from "typescript";
 import path from "path";
-import { OpenAPIV3 } from "openapi-types";
+import { OpenAPIV3, OpenAPIV3_1 } from "openapi-types";
 import * as cg from "./tscodegen";
 import generateServers, { defaultBaseUrl } from "./generateServers";
 import { Opts } from ".";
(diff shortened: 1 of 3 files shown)


src/codegen/generate.ts
```diff
@@ -695,7 +695,19 @@ export default class ApiGenerator {
...
     if (schema.type) {
-      // string, boolean, null, number
+      // string, boolean, null, number, array
+      if (Array.isArray(schema.type)) {
+        return factory.createUnionTypeNode(
+          schema.type.map((type) => {
+            if (type === null) return cg.keywordType.null;
+            if (type === "integer") return cg.keywordType.number;
+            if (type in cg.keywordType)
+              return cg.keywordType[type as keyof typeof cg.keywordType];
```
SimonEggert, line 705: Typecasting here because `schema` is still only of type `OpenAPIV3.SchemaObject`. Making that a union with `OpenAPIV3_1.SchemaObject` would be the cleaner way to go, but I don't know how much that entails.

Xiphe:
Hi @SimonEggert thank you for the contribution 🤙 This looks great already!

For reference, can you post a link to the spec here? Are these actually called "array types"? Feels very similar to a union to me.

Also I wonder if we can update the `OpenAPI`-Types so that we don't need the typecasting. Haven't touched that part though and am unsure about it. Would you mind checking if this is easily possible?

SimonEggert:
> For reference, can you post a link to the spec here? Are these actually called "array types"? Feels very similar to a union to me.

I edited the PR description with links to both the migration guide from OpenAPI and the JSON Schema docs because OpenAPI 3.1.0's aim was to achieve compatibility with the latter and that is where it is best documented.

> Also I wonder if we can update the OpenAPI-Types so that we don't need the typecasting. Haven't touched that part though and am unsure about it. Would you mind checking if this is easily possible?

I have tried to import and use both `OpenAPIV3` and `OpenAPIV3_1` types into the file, but didn't get it to work as there are more differences between the versions than just the changes of this PR. Furthermore some old features like `nullable: true` do not exist in the new version making it complicated. We would need to change multiple lines of code and add checks for V3 vs. V3_1 to account for the Type differences.

I might be able to look further into that, but maybe that would be in a wider scope than what the PR title suggests right now because it would include more general changes to support V3_1 (which would be pretty nice to have but would maybe require a bigger rewrite). I will let you know if I find a way in between 😉

src/codegen/generate.ts
```diff
@@ -21,6 +21,34 @@ type ContentType = "json" | "form" | "multipart";
...
+  | OpenAPIV3_1.ResponsesObject;
+type OpenAPIRequestBodyObject =
+  | OpenAPIV3.RequestBodyObject
+  | OpenAPIV3_1.RequestBodyObject;
+type OpenAPIMediaTypeObject =
+  | OpenAPIV3.MediaTypeObject
+  | OpenAPIV3_1.MediaTypeObject;
+type OpenAPIOperationObject =
+  | OpenAPIV3.OperationObject
+  | OpenAPIV3_1.OperationObject;
```
SimonEggert, line 50: This could well be extracted into a separate file for my taste, but all other types seem to be in here as well, so I left it for now.

SimonEggert:
@Xiphe

I have pushed another commit that used union types to support both V3 and V3_1. I had to make some more changes, but most of them are just type guards to make TypeScript happy. Let me know what you think 🙂

PS: "Ignore Whitespace"-diff recommended 😉

Xiphe:
Wow @SimonEggert thanks for taking the time to bring this in! LGTM!

Issue #562, opened 2024-01-30
[Feature Request] Plugin System


fimion:
Based on #293 and #341 and my own interests (I want to decorate the generated functions), I would like to formally open a feature request for a Plugin System for the oazapfts code generation. 

The initial points we should probably define are:

- plugin registration method
- primary points for plugin modification

I'm more than happy to lead the charge on this.

Xiphe:
Hi @fimion!
Thanks for bringing this up! I still think that a Plugin System would be a great benefit for this library. (What do you think @fgnass @jonkoops?)

As I'm basically in maintenance & PM mode for oazapfts currently I don't think I can be of much help but when you're interested in leading this I'd be more then happy to assist in design and review 👍

Xiphe:
My gut-feeling for this is, that the codebase is not in the greatest shape to be hooked into so this might bring some long needed refactors with it.
My personal hope would be that this would not only bring in a great feature to the outside but also causes the codebase to become cleaner 🤞
Or in other words: I would like to not just jam this in 😅

But before working on actual code, let's discuss the points you already brought up

fimion:
It's understandble to want to not just jam this in.

## Plugin Registration

I'm imagining from a generator cli end user standpoint that we would provide a command line argument that would let you point to multiple scripts that would be able to register at start up.

From a development perspective, I'd have these scripts need an object as default export that would have a predefined shape to them with potential methods attached that could hook into the various parts. We could expose a method to help with the type definitions as well as potentially do some run time checks? 

a plugin file would ultimately look something like this:

```ts
import { defineOazapftsPlugin } from "oazapfts/codegen";

export default defineOazapftsPlugin({
  // Various methods and or properties here
 });
 ```

## Initial Registration points

From an outside developer perspective (who has not fully studied the code), here are some initial places i could see allowing people to hook into stuff

- After downloading the oas file/before AST generation (access to the full OAS object)
- before converting a schema to a type (access to schema object, access to full AST object, access to full OAS object)
- after converting a schema to a type (access to schema object, access to schema-to-type AST object, access to full AST object, access to full OAS object)
- before converting an endpoint to a function (access to current endpoint object, access to full AST object, access to full OAS object)
- after converting an endpoint to a function (access to current endpoint object, access to endpoint-to-function AST object, access to full AST object, access to full OAS object)
- after AST generation is complete (access to full AST, access to full OAS object)
[...]

PR #572, opened 2024-02-11
[WIP] Code Generation Plugin System

fimion:
This is the first draft of a plugin system for the code generation part of Oazapfts (Discussion can be found in #562)


src/codegen/__future__/definePlugin.ts
```diff
@@ -0,0 +1,211 @@
...
+   */
+  afterASTGenerate?(
+    fullTree: ts.SourceFile,
+    document: OpenAPIV3.Document,
+  ): void | Promise<void>;
+
+  /**
+   * Runs after the TS file contents have been generated, but before writing to file.
+   * Can be async
+   * @param fileName - name of the file to write to (should this be a full path object?)
```
Xiphe, line 142: When we have the full path, yes. One could get the fileName when needed from this.

src/codegen/__future__/definePlugin.ts
```diff
@@ -0,0 +1,211 @@
...
+import ts from "typescript";
+import { Opts } from "../index";
+import minimist from "minimist";
+import assert from "node:assert";
+
+export type OazapftsCodegenPluginDefinition = {
+  /**
+   * Required identifier for this plugin
+   */
+  name: string;
```
Xiphe, line 12: I'd suggest to use the plugin object itself as an identifier. 
The `name` would still be useful for logging when needed. But might not be required.

src/codegen/__future__/definePlugin.ts
```diff
@@ -0,0 +1,211 @@
...
+
+export function defineOazapftsPlugin(
+  plugin: OazapftsCodegenPluginDefinition,
+): OazapftsCodegenPlugin {
+  // Do some checks to ensure that we have a valid plugin (if there are required params)
+  let isPlugin = true;
+  if (!plugin.name || plugin.name === "") {
+    console.warn("defineOazapftsPlugin: 'name' property is required!");
+    isPlugin = false;
+  }
```
Xiphe, line 197: Let's not do this for now. Given the name is only needed for logging, we can default to `unknown | anonymous` without hurting anyone badly. 

If we find we need validation let's throw an Error when these fail. Id argue there is no value in a `isPlugin=false` plugin.

Xiphe (approved):
This is looking great already!

Xiphe:
@fimion heads up I'm currently splitting the package: https://github.com/oazapfts/oazapfts/pull/576 
In order to prevent merge conflicts it might be wise to not work on big refactors until that one is merged. Let me know when you need something 👍

Xiphe:
The PR is ready to merge from my end, it just needs another pair of eyes and I think it would make sense when a few people could beta-test `oazapfts@6.0.0-beta.1` and `@oazapfts/runtime@1.0.0-beta.1`.

So when you like, a review and or a test would definitely help getting this out 🤙

src/codegen/__future__/definePlugin.ts
fimion: i've updated the type of this.

src/codegen/__future__/definePlugin.ts
fimion: I've made this not required

src/codegen/__future__/definePlugin.ts
fimion: i've removed the logic for this.

src/codegen/__future__/definePlugin.ts
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * Description for the plugin options.
+     */
+    description: string;
+    /**
+     * Validate if the options you have are valid or not.
+     * options will be considered valid by default if not validator plugin is provided.
+     * @param opts - the parsed options object.
+     */
+    validator?(opts: Opts): Boolean;
```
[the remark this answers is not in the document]
fimion: updated description to say that an error should be thrown. should i add a `@throws ValidationError` tsdoc thing?

Xiphe:
Thanks for incorporating my feedback!

Mentally iterating on this...

With the idea in the back of my head that I would like to put all oazapfts logic into a default plugin I wonder:

Should we try to create an architecture that...

1. allows plugins to set their priority on a given hook? 
2. allow plugins to cancel all other plugins on a given hook?

The Idea came from looking at the `afterWriteToFile`... it's a constructed scenario but what if there would be a plugin that writes the output directly to `#cloudStorage`. Ideally that plugin would want to run before the default implementation and then also declare that the original logic of writing to disk should be skipped.

So to support this, I'd suggest to use some sort of a priority setting on a hook rather then `before...` and `after...` also there would need to be a way to bail.

maybe we can leverage [tapable](https://www.npmjs.com/package/tapable) or a similar library? (I don't have experience with tapable, just was the first solid looking solution that I found)

An API for this approach will likely look more imperative, for example:

```ts
import { createPlugin } from "oazapfts";

export default createPlugin(({ apiGenerator }) => {
  apiGenerator.hooks.convertSchema.tap("just-use-any", () => {
    return ts.factory.createKeywordTypeNode(ts.SyntaxKind.AnyKeyword);
  });

  apiGenerator.hooks.write.tapPromise(
    "write-to-cloud",
    { stage: 0 },
    async ({ fileContents, fileName }) => {
      await myCloud.write(fileContents, fileName);
      return true;
    },
  );
});
```

---

Not feeling 100% confident with this. Might also be overkill 🤷 what do you think @fimion ?

Xiphe:
Cool! Thank you for the feedback! 

> If we are wanting to move the code generation process to be a default plugin, then we need to have a way to block cancelling certain steps. or we need to have a baseline set of functionality that always happens before breaking out the default functionality into a plugin itself. Or maybe not, I don't know. I think I'm having trouble thinking of a case where I would want to skip the OAS->AST generating steps, but that is probably just me having a limited perspective.

I also don't have a concrete scenario for this. My motivation for moving the core into a plugin is

1. I hope that it will make the core logic easier to maintain and to reason about
2. I hope this will enable use cases that we don't know about yet, while doing so 👆. Meaning a win-win scenario.

fimion:
@Xiphe I'm also looking at https://github.com/unjs/hookable as an option. It's typescript and is used by nuxt (I'm partial to vue, so i'm extremely biased here) but i think both of these could be good options, and the way they would get used would be very similar.

Would you prefer tapable or hookable?

Xiphe:
Skimming over it, hookable seems to have no support for manual ordering as tapable has with `stage`. Which would be better suited for a `beforeHook => nativeImplementation => afterHook` architecture.

Where tapable seems to order taps by `stage` which would allow us to tap the native implementation to `stage: 10` and then plugins might have finer control on wether they want to be the VERRY FIRST or rather somewhere in the middle.
_(Again: I have not used this, so I might be completely off in both: how this is supposed to work as well as how useful it is)_

Native typescript is a plus. Also I'm actually not a fan of the tapable ergonomics, it just seemed to cover the feature base we might need here and looking at the repo activity of tapable its either done or abandoned 🤷

I'd say let's not block ourselves with this decision, it seems to me as both point towards a similar direction which I like. I'd assume switching should not be a huge deal. 

---

That's all to say: I'm happy with both :)

Xiphe:
Hi @fimion 

I gave this another thought...

We might also just create the plugin interface as a collection of arrays with callbacks in it. 

```ts
type InitFn = (args: unknown) => unknown;
type ConvertSchemaFn = (args: unknown) => unknown;
type Oazapfts = {
  init: InitFn[];
  convertSchema: ConvertSchemaFn[];
};
```

A plugin could then just manipulate these arrays:

```ts
definePlugin((o: Oazapfts) => {
  // add to the end 
  o.init.push(myImplementation);
  // replace all existing
  o.convertSchema = [letMeHandleAllOfThis!];
};
```
  
I think this is the simplest solution that still provides a ton of flexibility.

fimion:
Okay @Xiphe I've made my first pass using the `tapable` package with proper types.

Xiphe:
Awesome! This looks like a great base to build upon! I'm not too sure if we actually need the `before` as we can set the `stage` when tapping in from a plugin but we can keep it for now. 

How would you suggest we continue from here?

fimion:
I think either this should be merged in for #584 or we should finish #584 then begin breaking out the default behavior into a plugin here. I'll also need to make a plugin loading system, which will change how the initial cli loading works. We can discuss that further if you'd like.

Xiphe:
Yeah I like the idea of merging with https://github.com/oazapfts/oazapfts/pull/584!

Xiphe:
Gonna merge this to main because:
- Everything is in the `__future__` folder
- This way GitHub understands @fimion as a contributor and CI will run automatically in future runs

Going forward I suggest we:
- iterate on the base plugin system from `main` (PRs to `main`)
- iterate on modularizing the source and make the plugin system work on `alpha` 
  - removal of the `ApiGenerator` class is a somewhat breaking change as people had been augmenting that one for custom behavior.
  - thats why I would like to release the finished plugin support along with removing the `ApiGenerator`

Issue #554, opened 2024-01-15
split up `generateApi` method


Xiphe:
The `generateApi` method over the time has become quite long. We might want to consider breaking it up into smaller pieces in order to increase maintainability 

ref https://github.com/oazapfts/oazapfts/pull/549

WildEgo:
This would be interesting, was wondering if this could offer the possibility to generate example snippets for the endpoints

Xiphe:
Hi @WildEgo That's a really interesting idea! Let's open a separate issue for that one, though. I don't think it's dependent on this refactoring ticket.

PR #584, opened 2024-02-23
Modularize

Xiphe:
Hand in hand with the [Plugin System](https://github.com/oazapfts/oazapfts/issues/562) i would like to refactor the codegen implementation to be more modular and follow a functional approach. 

Target state:
- [ ] much smaller, clearly defined units. - Most methods of `ApiGenerator` might be moved to their dedicated file.
- [ ] no functional changes
- [ ] extendability through [plugin system](https://github.com/oazapfts/oazapfts/issues/562) instead of monkey patching `ApiGenerator`

ref https://github.com/oazapfts/oazapfts/issues/554


Xiphe:
Moving to beta -> https://github.com/oazapfts/oazapfts/pull/805

PR #605, opened 2024-03-26
fix: support boolean schemas

emonadeo:
oazapfts codegen fails if the api contains [boolean json schema](https://json-schema.org/draft/2020-12/json-schema-core#section-4.3.2).

```jsonc
{
  // ...
  "schemas": {
    "Foo": {
      "type": "object",
      "required": ["bar"],
      "properties": {
        "bar": true
      },
      "additionalProperties": false
    }
  }
}
```

This PR adds support for these, mapping `true` to `any` and `false` to `never`.

I am not sure about how this should behave with "TrueEnums" (~~not sure what that is~~ i get it now, these are typescript enums), so please review with extra care, as I am not familiar with the codebase and don't understand every corner of it.

This is an upstream issue with `kogosoftwarellc/open-api`, which is also missing boolean schema support (I filed a PR to fix this at kogosoftwarellc/open-api#902).

### DIFF
--- a/packages/codegen/src/generate.ts
+++ b/packages/codegen/src/generate.ts
@@ -477,7 +477,7 @@ export default class ApiGenerator {
             type,
           }),
         );
-        return this.refs[$ref].base;
+        return typeReferenceNode.base;
       }
 
       if (ignoreDiscriminator) {
--- a/packages/codegen/src/generate.ts
+++ b/packages/codegen/src/generate.ts
@@ -457,12 +457,16 @@ export default class ApiGenerator {
 
     if (!this.refs[$ref]) {
       let schema = this.resolve<SchemaObject>(obj);
-      if (typeof schema !== "boolean" && ignoreDiscriminator) {
+
+      if (typeof schema === "boolean") {
+        return this.refs[$ref].base;
+      }
+
+      if (ignoreDiscriminator) {
         schema = _.cloneDeep(schema);
         delete schema.discriminator;
       }
-      const name =
-        (typeof schema !== "boolean" && schema.title) || getRefName($ref);
+      const name = schema.title || getRefName($ref);
       const identifier = toIdentifier(name, true);
 
       // When this is a true enum we can reference it directly,
--- a/packages/codegen/src/generate.ts
+++ b/packages/codegen/src/generate.ts
@@ -458,33 +458,12 @@ export default class ApiGenerator {
     if (!this.refs[$ref]) {
       let schema = this.resolve<SchemaObject>(obj);
 
-      if (typeof schema === "boolean") {
-        const name = getRefName($ref);
-        const identifier = toIdentifier(name, true);
-        const alias = this.getUniqueAlias(identifier);
-        const type = this.getTypeFromSchema(schema, undefined);
-        const typeReferenceNode = {
-          base: factory.createTypeReferenceNode(alias, undefined),
-          // `readOnly` and `writeOnly` are not applicable to boolean schemas
-          readOnly: undefined,
-          writeOnly: undefined,
-        };
-        this.refs[$ref] = typeReferenceNode;
-        this.aliases.push(
-          cg.createTypeAliasDeclaration({
-            modifiers: [cg.modifier.export],
-            name: alias,
-            type,
-          }),
-        );
-        return typeReferenceNode.base;
-      }
-
-      if (ignoreDiscriminator) {
+      if (typeof schema !== "boolean" && ignoreDiscriminator) {
         schema = _.cloneDeep(schema);
         delete schema.discriminator;
       }
-      const name = schema.title || getRefName($ref);
+      const name =
+        (typeof schema !== "boolean" && schema.title) || getRefName($ref);
       const identifier = toIdentifier(name, true);
 
       // When this is a true enum we can reference it directly,
--- a/packages/codegen/src/generate.ts
+++ b/packages/codegen/src/generate.ts
@@ -459,6 +459,24 @@ export default class ApiGenerator {
       let schema = this.resolve<SchemaObject>(obj);
 
       if (typeof schema === "boolean") {
+        const name = getRefName($ref);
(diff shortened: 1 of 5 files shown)


Xiphe:
Hi @emonadeo!

Thank you for the contribution and sorry for the late reply, I currently don't have much capacity for this project.

Could you help me wrap my head around why and where this is useful in the context of Open API Spec?
I don't want to imply, it's not useful. At the same time, as a maintainer of this I try to understand how people use the Spec and what they feed into the codegen and why.

emonadeo:
I personally don't use boolean json schema but others do and some codegen tools use it as well. Right now oazapfts just errors when trying to generate a client from such an openapi file.

packages/codegen/src/generate.ts
```diff
@@ -756,7 +776,11 @@ export default class ApiGenerator {
 
   isTrueEnum(schema: SchemaObject, name?: string): name is string {
     return Boolean(
-      schema.enum && this.opts.useEnumType && name && schema.type !== "boolean",
+      typeof schema !== "boolean" &&
+        schema.enum &&
+        this.opts.useEnumType &&
+        name &&
+        schema.type !== "boolean",
```
Xiphe, line 783: we have `schema.type !== "boolean"` two times in the conditional. can we remove this line?

packages/codegen/src/generate.ts
```diff
@@ -447,11 +457,12 @@ export default class ApiGenerator {
...
     if (!this.refs[$ref]) {
       let schema = this.resolve<SchemaObject>(obj);
-      if (ignoreDiscriminator) {
+      if (typeof schema !== "boolean" && ignoreDiscriminator) {
         schema = _.cloneDeep(schema);
         delete schema.discriminator;
       }
-      const name = schema.title || getRefName($ref);
+      const name =
+        (typeof schema !== "boolean" && schema.title) || getRefName($ref);
```
Xiphe, line 465: Can we do an early return in case the schema is bool? Would need to check what to return in that case.

packages/codegen/src/generate.ts
```diff
@@ -657,8 +676,9 @@ export default class ApiGenerator {
           isReference(childSchema) &&
           this.discriminatingSchemas.has(childSchema.$ref)
         ) {
-          const discriminatingSchema = this.resolve<SchemaObject>(childSchema);
-          const discriminator = discriminatingSchema.discriminator!;
+          const discriminatingSchema =
+            this.resolve<DiscriminatingSchemaObject>(childSchema);
+          const discriminator = discriminatingSchema.discriminator;
```
Xiphe, line 681: Thanks for taking care of this <3

packages/codegen/src/generate.ts
```diff
@@ -792,6 +816,13 @@ export default class ApiGenerator {
     with a new name adding a number
   */
   getTrueEnum(schema: SchemaObject, propName: string) {
+    if (typeof schema == "boolean") {
+      // HACK: this should never be thrown, since the only `getTrueEnum` call is
+      // behind an `isTrueEnum` check, which returns false for boolean schemas.
+      throw new Error(
+        "cannot get enum from boolean schema. schema must be an object",
+      );
```
Xiphe, line 824: Fair! Thanks for the clear comment! I'd personally not consider this a "HACK", it's just an additional guard that should never be invoked.

packages/codegen/src/generate.ts
```diff
@@ -792,6 +816,13 @@ export default class ApiGenerator {
     with a new name adding a number
   */
   getTrueEnum(schema: SchemaObject, propName: string) {
+    if (typeof schema == "boolean") {
```
Xiphe, line 819: Please use `===` :)

Xiphe (changes requested):
Solid work! I have some minor change requests 🤓

packages/codegen/src/generate.ts
emonadeo: Are you sure we can do this? `typeof schema !== "boolean"` and `schema.type !== "boolean"` are two different checks.

Xiphe: you're right! my bad

packages/codegen/src/generate.ts
emonadeo: Not sure how that would look like. The method technically just returns a `ts.TypeNode`, but it also some side effects, mutating `this.refs` and `this.aliases`.

Xiphe: Yeah, might be that this is not possible - will have a look

Xiphe: Added a commit - could you take a look?

Xiphe (approved):

packages/codegen/src/generate.ts
emonadeo: I added an (unfinished) test for referenced boolean schemas, but it errors:
```
TypeError: Cannot read properties of undefined (reading 'base')
 ❯ ApiGenerator.getRefAlias codegen/src/generate.ts:462:32
    460| 
    461|       if (typeof schema === "boolean") {
    462|         return this.refs[$ref].base;
       |                                ^
    463|       }
```

emonadeo: I pushed a fix (722bbde) but it repeats some code. Do you think we should refactor this a bit more?

Xiphe: Nice work! I think we should go back to what you initially had. Compared to the current solution it's way less code and repetition and it also actually works. 
Sorry for the back and forth, I'll add a change that keeps your great tests and moves the implementation back to what you initially proposed.

emonadeo: No worries. Didn't spend a lot of time on this at all.

Xiphe: Added it in. What do you think?

emonadeo: LGTM 👍🏼

PR #638, opened 2024-05-16
fix: cast booleans to string while forming multipart/form-data

### LINKED ISSUE #637: Bug: multipart should cast boolean values to string
`multipart` method of runtime should cast provided boolean values to strings before pushing them to FormData object, just as it does for numbers. This is as [per spec](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append) as well.

I think this was an omission while changing the logic [the last time](https://github.com/oazapfts/oazapfts/commit/b0058dee6183256bbff6c4f0562a8775305a0c11#diff-fac9cfbc478972c5ff1f5986c54a8fc43cc87a3f50e9fd35acfae178bb966fd5).

ekarademir:
Closes #637

### DIFF
--- a/packages/runtime/src/index.test.ts
+++ b/packages/runtime/src/index.test.ts
@@ -125,11 +125,17 @@ describe("request", () => {
   
   it("casts numbers and booleans to strings while forming multipart/form-data", () => {
     const multipartRequest = oazapfts.multipart({
-      body: { booleanValue: true },
+      body: { 
+        numberValue: 42,
+        booleanValue: true 
+      },
     });
 
     expect(multipartRequest.body.get('booleanValue')).toBe(
       'true'
+    );
+    expect(multipartRequest.body.get('numberValue')).toBe(
+      '42'
     );
   });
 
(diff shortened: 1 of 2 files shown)


Xiphe:
Thank you for taking care of this and sorry I wasn't able to respond sooner.

I'm a little confused right now

According to [mdn](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append) We're supposed to not case to string and from what I understand, FormData would do that for us.

```ts
formData.append("name", true);
formData.append("name", 72);
formData.getAll("name"); // ["true", "72"]
```

So from what I understand the issue is that we're falling into the else path and packing the boolean into a JSON blob 

https://github.com/oazapfts/oazapfts/blob/6cdc432c3fe515055b1980a11aac7f00bd172c6f/packages/runtime/src/runtime.ts#L153-L162

which is clearly unwanted and a bug.

That said, I'm puzzled because I cant' reconstruct why we've been casting Numbers to Stings in the first place.
https://github.com/oazapfts/oazapfts/pull/623/files did not do this - but the current state (https://github.com/oazapfts/oazapfts/commit/810c3fa97c6d77e1f7f08e291bd04c827c86635b) isn't using that Solution and I'm trying to find out why.

Xiphe:
Ok, typescript does not like it when we append booleans or numbers :)

LGTM

Xiphe:
moved to https://github.com/oazapfts/oazapfts/pull/680 because I didn't know how to append commits here :)

PR #670, opened 2024-08-02
feat: added enum array support

VladNcode:
An array of enums is handled incorrectly when the useEnumType option is enabled.


Xiphe:
Great work! Thank you

PR #672, opened 2024-09-20
fix: ignore undefined discriminator mapping

cameronevans:
Ran into an issue trying to generate types for a schema that had a discriminator defined without mapping. The missing mapping was caused by the generated OpenAPI spec having poorly parsed inheritance.

The following will result in `export type MyAlias = ;`, with the conditional change it will fall back to `export type MyAlias = {};`
```json
 "MyAlias": {
        "required": ["@type"],
        "type": "object",
        "properties": {},
        "discriminator": { "propertyName": "@type" }
      },
```

### DIFF
--- a/demo/api.ts
+++ b/demo/api.ts
@@ -129,6 +129,7 @@ export type Issue542 = {
   "media-protocol"?: "hls" | "mss" | "dash";
   "dashed-property"?: string;
 };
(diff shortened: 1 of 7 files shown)


Xiphe:
Thanks for the PR! And sorry that I wasn't available last year 🤙

PR #681, opened 2025-02-12
fix: pass required properties to children in `allOf` schemas

12:
According to the OpenAPI spec, annotating an `allOf` schema with a `required` property is valid. I don't believe this package was fully supporting that use case, as we were observing that with the following spec:

```yml
    getPerson:
      type: object
      allOf:
        - type: object
          properties:
            firstName:
              type: string
            secondName:
              type: string
      required:
        - firstName
```

We were getting this result:

```
{
    firstName?: string;
    secondName?: string;
}
```

When it should in fact be:

```
{
    firstName: string;
    secondName?: string;
}
```

I think we can fix this by simply passing the required property on the parent through with the child schema, and has in fact been working for us (as an applied `patch-package`) for some time now. But also, if this is completely wrong, please do let me know.

### DIFF
--- a/packages/codegen/src/generate.test.ts
+++ b/packages/codegen/src/generate.test.ts
@@ -42,6 +42,39 @@ describe("content types", () => {
   });
 });
 
+describe("getBaseTypeFromSchema", () => {
+  it("should generate type with required properties when extending a schema with the 'allOf' operator", () => {
+    const generator = new ApiGenerator({} as unknown as OpenAPIV3.Document);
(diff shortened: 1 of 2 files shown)


Xiphe:
Nice! 

I guess that would also be true for anyOf and oneOf types, right?

Xiphe:
Could you give me a reference to where in the spec this behavior is defined?

Xiphe:
Got it! Thanks for explaining.

Issue #674, opened 2024-09-26
Enum Discriminator Issue


randdusing:
Howdy,

useEnumType doesn't appear to affect discriminators. I played around with this test: https://github.com/oazapfts/oazapfts/blob/77d47051d12423c959f4f5cd14b155119471cebd/packages/codegen/src/index.test.ts#L70 and the json that it's consuming: https://github.com/oazapfts/oazapfts/blob/77d47051d12423c959f4f5cd14b155119471cebd/packages/codegen/src/__fixtures__/allOf.json#L109

By default, the resulting code looks like this:
```
export type PetBase = { petType: string; };
export type Dog = { petType: "dog"; } & PetBase & { bark?: string; };
export type Cat = { petType: "Cat"; } & PetBase & { name?: string; };
export type Lizard = { petType: "Lizard"; } & PetBase & { lovesRocks?: boolean; };
export type Pet = Dog | Cat | Lizard;
```

Then I modified allOfJson so that petType was an enum, like this:
```
"Pet": {
  "type": "object",
  "properties": {
    "petType": {
      "type": "string",
      "enum": ["dog", "Cat", "Lizard"]
    }
  },
  "required": ["petType"],
  "discriminator": {
    "propertyName": "petType",
    "mapping": {
      "dog": "#/components/schemas/Dog"
    }
  }
}
```

And then re-ran the test with useEnumType set to true, which resulted in this code:
```
export type PetBase = { petType: PetType; };
export type Dog = { petType: "dog"; } & PetBase & { bark?: string; };
export type Cat = { petType: "Cat"; } & PetBase & { name?: string; };
export type Lizard = { petType: "Lizard"; } & PetBase & { lovesRocks?: boolean; };
export type Pet = Dog | Cat | Lizard;
export enum PetType { Dog = "dog", Cat = "Cat", Lizard = "Lizard" } 
```

When I use the Dog class, TypeScript determines it to be "never" since the petType property's "dog" | PetType results in never. Although I don't completely understand why this is the case since PetType.Dog === "dog"

This is what I think it should look like when using useEnumType.
```
export type PetBase = { petType: PetType; };
export type Dog = { petType: PetType.Dog; } & PetBase & { bark?: string; };
export type Cat = { petType:  PetType.Cat; } & PetBase & { name?: string; };
export type Lizard = { petType:  PetType.Lizard; } & PetBase & { lovesRocks?: boolean; };
export type Pet = Dog | Cat | Lizard;
[...]

Issue #611, opened 2024-04-04
Self-generated API file with hard-coded "types"


kmirojo:
Hi everyone,
I'm generating an API `typescript` file from an **Open API** `.yml` file, but something is happening.
My `.yml` file looks like this:
```yaml
components:
  schemas:
    ItemPropertiesBase:
      type: object
      properties:
        type:
          type: string

    ItemProperties:
      allOf:
        - $ref: '#/components/schemas/ItemPropertiesBase'
        - type: object
          properties:
            prop1:
              type: boolean
            prop2:
              type: number
            prop3:
              type: number
```

but when I generate the code, the `type` prop from the `ItemPropertiesBase` is **hard-coded**, even though I only assigned a type, but not a value:

```ts
export type ItemPropertiesBase = {
    "type": string;
};

export type ItemProperties = {
    "type": "ItemProperties"; 
} & ItemPropertiesBase & {
    prop1: boolean;
    prop2: number;
    prop3: number;
};
```

The issue with this is that I need the `type` to be a `string`, but not a specific value, because otherwise `typescript` won't allow me to have a different `type`,
Is there a way to change this?

Xiphe:
Hi @kmirojo 

Seems as if you're running into compatibility issues with our "discriminating schemas" (see https://github.com/oazapfts/oazapfts/issues/422). Do you use a `discriminator` in one of your real-world types?

And can you please confirm, the expected outcome would be something along the lines of 

```ts
export type ItemPropertiesBase = {
    "type": string;
};

export type ItemProperties = ItemPropertiesBase & {
    prop1: boolean;
    prop2: number;
    prop3: number;
};
```

randdusing:
I'm running into a similar issue. The eventSourceType defined on the child classes isn't a valid value in the EventSourceType enum. Is there a way to remove the hardcoded eventSourceType that's defined on the child classes?

```javascript
export enum EventSourceType {
    A = "A",
    B = "B",
}

export type EventSourceBase = {
    eventSourceType: EventSourceType;
    eventSourceId: number;
};
export type AEventSource = {
    eventSourceType: "AEventSource";
} & EventSourceBase & {
    date: string;
    amount: number;
};
export type BEventSource = {
    eventSourceType: "BEventSource";
} & EventSourceBase & {
    date: string;
    amountNear: number;
    amountFar: number;
};
```

Xiphe:
So this is related to https://github.com/oazapfts/oazapfts/issues/674 right?

[...]

PR #741, opened 2025-07-15
Fix issue with enum discriminators

issy:
When generating code from a schema like this
```
"Pet": {
  "type": "object",
  "properties": {
    "petType": {
      "type": "string"
    }
  },
  "required": ["petType"],
  "discriminator": {
    "propertyName": "petType",
    "mapping": {
      "dog": "#/components/schemas/Dog",
      "poodle": "#/components/schemas/Dog",
      "cat": "#/components/schemas/Cat"
    }
  }
}
```
I would expect to see the following TypeScript output
```ts
export type PetBase = { petType: string };
export type Dog = { petType: "dog" | "poodle" } & PetBase & { bark?: string };
export type Cat = { petType: "cat" } & PetBase & { name?: string };
export type Pet = Dog | Cat;
```
However, `oazapfts` generates the following instead
```ts
export type PetBase = { petType: string };
export type Dog = { petType: "dog" } & PetBase & { bark?: string };
export type Cat = { petType: "Cat" } & PetBase & { name?: string };
export type Lizard = { petType: "Lizard" } & PetBase & { lovesRocks?: boolean };
export type Pet = Dog | Cat | Lizard;
```
There is a problem with this - `Dog` can't be picked by the value `'poodle'`. It picks the "first" mapping entry instead of using all mapping entries. This has caused issues in my workflow, as the generated code for our web app ends up having incomplete type information. This issue has been logged in https://github.com/oazapfts/oazapfts/issues/674 and https://github.com/oazapfts/oazapfts/issues/611

In my case, my schema has 3 mapping entries pointing to the same schema. I'm sure you can see how this would cause issues.

Anyway, this PR aims to resolve that.

### DIFF
--- a/packages/codegen/src/index.test.ts
+++ b/packages/codegen/src/index.test.ts
@@ -11,7 +11,7 @@ const demoFolder = path.join(rootFolder, "demo");
 /**
  * Generate an API from a relative path and convert it into a single line.
  */
-async function generate(file: string, opts: Opts = {}): Promise<string> {
+async function generate(file: string, opts: Opts = {}) {
   const src = await generateSource(file, opts);
   const error = await checkForTypeErrors(src);
   expect(error).toBeUndefined();
--- a/packages/codegen/src/index.test.ts
+++ b/packages/codegen/src/index.test.ts
@@ -60,8 +60,12 @@ describe("generateSource", () => {
     expect(src).toContain(`export type Option = ("one" | "two" | "three")[];`);
   });
 
-  it(async () => {
-    const src = await generate(__dirname + "/__fixtures__/allOf.json");
+  describe("discriminator mappings with allOf", () => {
+    let src: string;
+
+    beforeAll(async () => {
+      src = await generate(__dirname + "/__fixtures__/allOf.json");
+    });
 
     it("should handle properties both inside and outside of allOf", async () => {
       expect(src).toContain(
--- a/packages/codegen/src/index.test.ts
+++ b/packages/codegen/src/index.test.ts
@@ -72,7 +72,10 @@ describe("generateSource", () => {
     expect(src).toContain("export type PetBase = { petType: string; };");
     expect(src).toContain("export type Pet = Dog | Cat | Lizard;");
     expect(src).toContain(
-      'export type Dog = { petType: "dog"; } & PetBase & { bark?: string; };',
+      'export type Cat = { petType: "cat"; } & PetBase & { name?: string; };',
+    );
+    expect(src).toContain(
+      'export type Dog = { petType: "dog" | "poodle"; } & PetBase & { bark?: string; };',
     );
     expect(src).toContain(
       'export type Lizard = { petType: "Lizard"; } & PetBase & { lovesRocks?: boolean; };',
--- a/packages/codegen/src/index.test.ts
+++ b/packages/codegen/src/index.test.ts
@@ -1,6 +1,6 @@
-import { describe, it, expect, beforeAll, assert } from "vitest";
+import { beforeAll, describe, expect, it } from "vitest";
 import * as path from "node:path";
-import { Opts, generateSource } from "./index";
+import { generateSource, Opts } from "./index";
 import { readFile } from "node:fs/promises";
 import { createProject, ts } from "@ts-morph/bootstrap";
 import { ScriptTarget } from "typescript";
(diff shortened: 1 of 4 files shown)


packages/codegen/src/index.test.ts
```diff
@@ -60,23 +60,28 @@ describe("generateSource", () => {
     expect(src).toContain(`export type Option = ("one" | "two" | "three")[];`);
   });
 
-  it("should handle properties both inside and outside of allOf", async () => {
+  it(async () => {
```
issy, line 63: Just nested these to deduplicate setup.

packages/codegen/src/index.test.ts
```diff
@@ -60,23 +60,28 @@ describe("generateSource", () => {
...
+      );
+    });
+
+    it("should support discriminator used in conjunction with allOf", async () => {
+      expect(src).toContain("export type PetBase = { petType: string; };");
+      expect(src).toContain(
+        'export type Cat = { petType: "cat"; } & PetBase & { name?: string; };',
+      );
+      expect(src).toContain(
+        'export type Dog = { petType: "dog" | "poodle"; } & PetBase & { bark?: string; };',
```
issy, line 78: When there are more than 1 mapping matches, we get a string union

packages/codegen/src/index.test.ts
```diff
@@ -60,23 +60,28 @@ describe("generateSource", () => {
...
+      );
+      expect(src).toContain(
+        'export type Lizard = { petType: "Lizard"; } & PetBase & { lovesRocks?: boolean; };',
+      );
+      expect(src).toContain("export type Pet = Dog | Cat | Lizard;");
```
issy, line 83: Deduplicated members here now.

packages/codegen/src/index.test.ts
```diff
@@ -11,7 +11,7 @@ const demoFolder = path.join(rootFolder, "demo");
 /**
  * Generate an API from a relative path and convert it into a single line.
  */
-async function generate(file: string, opts: Opts = {}) {
+async function generate(file: string, opts: Opts = {}): Promise<string> {
```
Xiphe, line 14: nit: Let's use implicit return type

packages/codegen/src/index.test.ts
Xiphe: Is it a common patter to nest `it` calls? Would have expected to use `describe` + `beforeEach|beforeAll` here.

packages/codegen/src/index.test.ts
Xiphe: I read this as "Prior to the fix, there were duplication here". But we had the same assertion in line 73 before. 

So how is this test validating the fix, or something else?

Xiphe: ...assuming that the duplication was added along with the fix, right? So Current version doesn't have duplication but with the change from `find` to `filter` it was added, then mitigated using `_.uniq` so that this assertion remains valid, right?

Xiphe (commented):
Hey @issy 

Thank you so much for taking the time to dig into this, really appreciated!

I have two nitty comments, you're invited to fix those or I can also quickly add them with the merge. 

I don't really see how this fixes https://github.com/oazapfts/oazapfts/issues/674, thought that issue was more about using the actual Enum type instead of literals when discriminating. When this is actually covered by this change I'd love to have a test-case for it.

 Also I don't see how this is fixing https://github.com/oazapfts/oazapfts/issues/611 as the `type` (or `petType` in this example) is still present - which is expected in this example using explicit discriminators but in the examples of the issue I read it as it's not expected as there is no explicit discriminator. Same here, when this change fixes the issue, I would love to see a test-case to proove it.
 
 In any case, this fixes the issue you're describing in the PR - happy to bring this in soon.

issy:
> I don't really see how this fixes #674, thought that issue was more about using the actual Enum type instead of literals when discriminating. When this is actually covered by this change I'd love to have a test-case for it.

@Xiphe correct, sort of. #674 is mostly about the actual Enum type (which this PR does not fix), but _part of_ that issue is the actual contents of the union, which this PR _does_ fix.

> Also I don't see how this is fixing #611 as the `type` (or `petType` in this example) is still present - which is expected in this example using explicit discriminators but in the examples of the issue I read it as it's not expected as there is no explicit discriminator. Same here, when this change fixes the issue, I would love to see a test-case to proove it.

Apologies for poorly referencing this issue. It seems I was actually referring more explicitly to issues brought up in the _comments_ of #611. The initial description in #611 doesn't elaborate on whether or not a discriminator is used, so I don't know if this will actually resolve that issue (I suspect a discriminator _is_ used, just going by the provided code output).

Furthermore, I'm not entirely sure if `Lizard` should still be included in the union, as it is not explicitly mentioned as a discriminator. I've kept it for now, as I suspect there are implementations in the wild that rely on this behaviour. I'm no OpenAPI expert, so maybe there is documentation explicitly stating that this discriminator should be implicitly included, as it is now.

I believe I've addressed your PR comments. Let me know if there's anything else!

Xiphe (approved):
LGTM :turtle:

